PR #9, opened 2020-06-04
Documentation mentions build.py

ashishmohite:
## Description
The documentation mentions `build.py` but the file is named `lambda.py`
The name `lambda.py` feels misleading as it feels like it's the source code of lambda rather than the helper script

## Motivation and Context
NA
## Breaking Changes
NA
## How Has This Been Tested?
NA


antonbabenko:
Thanks for opening this PR!

I've updated the docs. `lambda.py` is the correct name, because the script does several things and not just "build".

PR #13, opened 2020-06-09
feat: Reliable passing build plan by a separate file + minor changes

ahlinc:
## Description
* Added passing of build plan by a separate file.
  Also in an archive_plan resource the build plan data are seen.
* Added docker_pip_cache variable to speedup a python requirements repeatable installations.
* Renamed: lambda.py -> package.py
  *It seems that lambda.py is a confusable name and we have the
  package.tf file so it's better to name the helper as package.py
  to make it associated as a companion script for the package.tf file.*


variables.tf
```diff
@@ -481,3 +481,9 @@ variable "docker_with_ssh_agent" {
   type        = bool
   default     = false
 }
+
+variable "docker_pip_cache" {
+  description = "Whether to mount a shared pip cache folder into docker environment or not"
+  type    = any
```
antonbabenko, line 487: `type = string` is better, right?

package.tf
```diff
@@ -14,20 +14,32 @@ data "external" "archive_prepare" {
...
-    hash_extra_paths = jsonencode(["${path.module}/lambda.py"])
+    hash_extra_paths = jsonencode(["${path.module}/package.py"])
   }
 }
 
+# This transitive resource used as a bridge between a state stored
+# in a Terraform plan and a call of a build command on the apply stage
+# to transfer a noticeable amount of data
+resource "local_file" "archive_plan" {
+  count = var.create && var.create_package ? 1 : 0
```
antonbabenko, line 35: Please run `pre-commit run -a` to fix the formatting of all files.

package.tf
```diff
@@ -3,7 +3,7 @@
 data "external" "archive_prepare" {
   count = var.create && var.create_package ? 1 : 0
 
-  program     = ["python", "${path.module}/lambda.py", "prepare"]
+  program     = ["python3", "${path.module}/package.py", "prepare"]
```
antonbabenko, line 6: I agree with the rename into `package.py`, please update README where it is mentioned, and also update Python required version there.

variables.tf
ahlinc: It's not, the variable may be used in two ways:
* docker_pip_cache = true - creates a cache folder inside of $artifacts_dir/cache/pip
* docker_pip_cache = "/tmp/some_folder_path" - it's possibility to put the cache folder somewhere outside of $artifacts_dir and don't have an issue with incorrect permissions due to the folder usage inside of a docker container which runs by default with a root user permissions.

antonbabenko: Right, makes sense.

package.tf
ahlinc: Done, a corresponding commit was fixed up.

package.tf
ahlinc: Done.

Issue #22, opened 2020-06-12
Feature Request: Ability to set the retention length on CloudWatch Log Group


tkalus:
**Feature Request for provisioning the expected CloudWatch Log Group in TF, allowing for a `retention_in_days` to be set.**

Something like:
**`logs.tf`**
```
resource "aws_cloudwatch_log_group" "lambda" {
  count = var.attach_cloudwatch_logs_policy && var.lambda_at_edge ? 0 : 1
  name = "/aws/lambda/${var.function_name}"
  retention_in_days = var.cloudwatch_logs_retention
}

resource "aws_cloudwatch_log_group" "lambda-edge" {
  count = var.attach_cloudwatch_logs_policy && var.lambda_at_edge ? 1 : 0
  name = "/aws/lambda/us-east-1.${var.function_name}"
  retention_in_days = var.cloudwatch_logs_retention
}
```
...with a complementary flip of the `local.log_group_arns` and removal of the `logs:CreateLogStream` perms.

On the minus-side, `retention_in_days` needs a value to be set, but defaulting `var.cloudwatch_logs_retention` to `3653` more-or-less matches the current behavior of "Never".

On the positive side, a module `output` detailing the Log Group makes it easier to attach a subscriber and route the logs centrally.

Thoughts?

antonbabenko:
Thanks for opening this issue. I will complete your PR once AWS IAM service is back (as of now it has issues with creating IAM roles).

Andrey9kin:
Nice! Was about to submit this as an issue but apparently it is already done

PR #23, opened 2020-06-12
Add provisioning for expected Lambda's CloudWatch Log Group

tkalus:
Closes terraform-aws-modules/terraform-aws-lambda#22

## Description
- Adds ability to set a retention time for the Lambda's CloudWatch Logs.
- Adds the Log Group to the outputs to allow for easy-attach of a
  subscriber.
- Retains the default `Never Expire` (days == 0) behavior and logs are
  not created if `create_role` or `attach_cloudwatch_logs_policy` are false;
  expectation is the CWLogGroup would be handled externally in that case.

## Motivation and Context
Would like to _not_ pay to retain old, crufty CloudWatch Logs
Would also like an easy way to attach a consumer to my Lambda's CWE Log Group.

## Breaking Changes
Not breaking; retains existing, expected behavior.

## How Has This Been Tested?
Used the examples/simple tf in this repo and twiddled some bits around a few times, changing the retention and `false`'ing variables like `attach_cloudwatch_logs_policy`.


antonbabenko:
https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/tag/v1.7.0

PR #28, opened 2020-06-16
fix: overlooked env dumping on DEBUG3

ahlinc:
Also removed examples/fixtures/dirtree folder with demo files, there is a script for it.
I'd suggest to make a minor version change for this.

### DIFF
--- a/package.py
+++ b/package.py
@@ -79,6 +79,11 @@ def configure_logging(use_tf_stderr=False):
     log.setLevel(logging.INFO)
 
 
+def dump_env():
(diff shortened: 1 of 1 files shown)


antonbabenko:
I know about the script. Here is the explanation why dirtree is there :) There are examples, which always should be possible to execute with the same commands (terraform init, plan, apply) across all terraform-aws-modules repositories. Either change example and remove reference of `dirtree` there (which is bad, because we want to test with complex dirtree, too), or leave `dirtree` dirs as is (even if it does not look so nice).

PR #30, opened 2020-06-16
fix: Fixed patterns applying

ahlinc:
8f0c5e6 *package.py - allow heredoc commands, don't apply patterns for explicitly specified single file*
94ec6bb *package.py - Fixed patterns applying without specified commands*


antonbabenko:
[v1.12.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.12.0) has been released.

PR #50, opened 2020-08-25
Fix: os xcode python interpreter

ahlinc:
Added workaround for OSX XCode python interpreter, but still doesn't know how to reproduce on OSX a situation when OSX XCode python interpreter becomes a main interpreter observable by terraform.
Also added an exception that better explains the issue with missing python interpreter version.


antonbabenko:
Thanks!

[v1.21.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.21.0) has been just released.

PR #52, opened 2020-08-26
fix: take build_plan into account when computing hash

ahlinc:
We need to take a build_plan into account when computing hash to properly trigger rebuilding.
The bug was captured on an isolated layer building and deployment.

### DIFF
--- a/package.py
+++ b/package.py
@@ -1047,6 +1047,7 @@ def prepare_command(args):
     hash_extra_paths = [p.format(path=tf_paths) for p in hash_extra_paths]
 
     content_hash = bpm.hash(hash_extra_paths)
+    content_hash.update(json.dumps(build_plan, sort_keys=True).encode())
     content_hash.update(runtime.encode())
     content_hash.update(hash_extra.encode())
     content_hash = content_hash.hexdigest()


package.py
```diff
@@ -1047,6 +1047,7 @@ def prepare_command(args):
     hash_extra_paths = [p.format(path=tf_paths) for p in hash_extra_paths]
 
     content_hash = bpm.hash(hash_extra_paths)
+    content_hash.update(json.dumps(build_plan, sort_keys=True).encode())
```
antonbabenko, line 1050: Thinking aloud... :)

`build_plan` contains _path_ which may lead to different `content_hash` depending on where this code is running. Is this ok?

Will it trigger new redeployment if running from different machines?

ahlinc: Good questions.
I need to check all of them first.
Will tell when I'll be ready later.

ahlinc: You are right by all questions, the whole build_plan can't be taken for hash. But there is a situation when valuable modifications in the build plan doesn't lead to a layer rebuilding and redeployment when a zip file already exists.
So there is the need to change strategy and take into account all important parts of build plan except of paths because
paths already got into account indirectly by files content and that seems should be enough. Such change requires significant code refactoring in the package.py

ahlinc: I think we need to close this PR as unacceptable and open an issue for the problem.

antonbabenko: What if we let the user specify/update `hash_extra` value for that? For example, `hash_extra` can take into account the content of the file using Terraform's function like [filemd5](https://www.terraform.io/docs/configuration/functions/filemd5.html)?

ahlinc: We introduced `hash_extra` as a temporary measure due to limitations of current package.py implementation and I thought that we will remove it in future when the implementation will be improved.
If there is a way how the user can refer to a file content from itself in the hash_extra this may work but this will look as a very strange hack.

ahlinc: I've opened an issue so we can decline this PR.

PR #58, opened 2020-09-14
feat: Added support for policy_jsons (list of strings)

### LINKED ISSUE #56: Add an explicit requirement on 'hashicorp/external'
`required_providers` does not list `external` and thus `terraform init` complains.

```
...
Initializing provider plugins...
- Checking for available provider plugins...
- Downloading plugin for provider "external" (hashicorp/external) 1.2.0...
- Downloading plugin for provider "aws" (hashicorp/aws) 2.70.0...

The following providers do not have any version constraints in configuration,
so the latest version was installed.

To prevent automatic upgrades to new major versions that may contain breaking
changes, it is recommended to add version = "..." constraints to the
corresponding provider blocks in configuration, with the constraint strings
suggested below.

* provider.external: version = "~> 1.2"

Terraform has been successfully initialized!
...
```

### LINKED ISSUE #57: Allow passing of multiple JSON-Policies
Hello,

we have a use case where we have the need to pass multiple rendered policy data to the module. However, currently, as far as i can see, only one policy document can be parsed. An implementation of an extension could look like this:

To stay backward-compatible, a new variable/resource is proposed.
```
variables:
  additional_jsons:
    type: list
    default: []
  attach_policies_json:
    type: bool (even list of bools thinkable)
    default: false (or [])
```

implementation example for one boolean value (cant make a pull request as i described in the issue before):

```
resource "aws_iam_policy" "additional_jsons" {
  count = local.create_role && var.attach_policies_json ? length(var.additional_jsons) : 0

  name   = "${var.function_name}-${count.index}"
  policy = var.additional_jsons[count.index]
}

resource "aws_iam_policy_attachment" "additional_jsons" {
  count = local.create_role && var.attach_policy_json ? length(var.additional_jsons) : 0

  name       = "${var.function_name}-${count.index}"
  roles      = [aws_iam_role.lambda[0].name]
  policy_arn = aws_iam_policy.additional_json[count.index].arn
}
```

Are there any things preventing this? In our current situation, because of dependency resolving problems, we cannot use your policy_statements variable (which would be much nicer but it is as it is) and dont want to create X combinations of permission documents for each single permission we want to allow.

antonbabenko:
Fixes #57
Fixes #56

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,10 @@
 repos:
   - repo: git://github.com/antonbabenko/pre-commit-terraform
-    rev: v1.31.0
+    rev: v1.39.0
     hooks:
       - id: terraform_fmt
       - id: terraform_docs
   - repo: git://github.com/pre-commit/pre-commit-hooks
-    rev: v3.1.0
+    rev: v3.2.0
     hooks:
       - id: check-merge-conflict
--- a/README.md
+++ b/README.md
@@ -240,12 +240,13 @@ module "vpc" {
 
 ## Additional IAM policies for Lambda Functions
 
-There are 4 supported ways to attach IAM policies to IAM role used by Lambda Function:
+There are 5 supported ways to attach IAM policies to IAM role used by Lambda Function:
 
-1. `policy_json` - set as JSON string or heredoc, when `attach_policy_json = true`.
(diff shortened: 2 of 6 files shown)


antonbabenko:
[v1.23.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.23.0) has been just released.

PR #71, opened 2020-11-17
feat: Updated range of supported versions of Terraform and providers

### LINKED ISSUE #70: Allow the module to use terraform-provider-random v3
Hi all,

The random provider 3.0 is out now for more than a month.
At the moment is not possible for any state using this module, directly or inderectly via another module, to update to the new version.

This is the changelog: 
- https://github.com/hashicorp/terraform-provider-random/blob/master/CHANGELOG.md#300-october-09-2020

Since this module does not support terraform 0.11 or the b64 setting, I don't think the extension to support 3.x should be an issue.

antonbabenko:
Closes #70

### DIFF
--- a/README.md
+++ b/README.md
@@ -556,21 +556,21 @@ Q4: What does this error mean - `"We currently do not support adding policies fo
 
 | Name | Version |
 |------|---------|
-| terraform | >= 0.12.6, < 0.14 |
-| aws | >= 2.67, < 4.0 |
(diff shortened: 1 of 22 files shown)


antonbabenko:
[v1.28.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.28.0) has been just released.

PR #76, opened 2020-11-23
fix: Fixed CodeDeploy hooks

### LINKED ISSUE #75: Deploy fails on invalid generated AppSpec file if a single hook is used
I'm using Terraform 0.13 (which might be important for this one, because it deals with `null`?)

I include a `deploy` configuration that looks something like this:

```hcl
module "deploy" {
  count = var.codedeploy ? 1 : 0
  source  = "terraform-aws-modules/lambda/aws//modules/deploy"
  version = "~> 1.29"

  # [...]
  before_allow_traffic_hook_arn = "some_lambda_function_name"
}
```

The only really notable thing about the config is that it has a before_allow_traffic hook, but no **after**_allow_traffic hook. This results in a deploy script that has an invalid AppSpec file.

When you apply a config like the above, inside `deploy_script_<hash>.txt`, there's the `aws deploy create-deployment` command. And that has a `--revision` that contains an AppSpec content that decodes to (phew):

```js
{
  "Hooks": [
    {
      "AfterAllowTraffic": null
    },
    {
      "BeforeAllowTraffic": "some_lambda_function_name"
    }
  ],
  "Resources": [
  // [...]
```

When this is sent to AWS, the result is a deployment failure: `The Lambda function name cannot be null or empty`. I've confirmed this is referring to the `AfterAllowTraffic` hook, because removing it allows the create-deployment command to succeed.

```json
"errorInformation": {
    "code": "INVALID_LAMBDA_CONFIGURATION",
    "message": "The deployment failed because the AppSpec file that specifies the AWS Lambda deployment configuration is missing or has an invalid configuration. The Lambda function name cannot be null or empty."
},
```

---

I think a good fix might be to change map(k, v) into map(k, v) if v != null at https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/dc6d2485e765b398b4e87fe15994b96dda56e22d/modules/deploy/main.tf#L28

antonbabenko:
Resolves #75

### DIFF
--- a/modules/deploy/main.tf
+++ b/modules/deploy/main.tf
@@ -25,7 +25,7 @@ locals {
     Hooks = [for k, v in zipmap(["BeforeAllowTraffic", "AfterAllowTraffic"], [
       var.before_allow_traffic_hook_arn != "" ? var.before_allow_traffic_hook_arn : null,
       var.after_allow_traffic_hook_arn != "" ? var.after_allow_traffic_hook_arn : null
-    ]) : map(k, v)]
+    ]) : map(k, v) if v != null]
   } : {})
 
   appspec_content = replace(jsonencode(local.appspec), "\"", "\\\"")


antonbabenko:
[v1.30.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.30.0) has been just released.

PR #84, opened 2020-12-11
docs: update description of hook vars, note naming expectations of default policy

### LINKED ISSUE #77: deploy - for default role, hook lambda names must begin with `CodeDeployHook_`
Terraform version: 0.13.5
`deploy` module version: 1.30.0

I use a configuration where I leave `create_codedeploy_role` at its default `true`, and rely on the module creating the role for me (I don't specify any `codedeploy_role_name`). I also pass in an `before_allow_traffic_hook_arn`. (All are needed to reproduce this bug)

When I went to deploy, the created role didn't seem to have sufficient permissions. The error was:

```js
{
    "deploymentInfo": {
        // [...]
        "status": "Failed",
        "errorInformation": {
            "code": "IAM_ROLE_PERMISSIONS",
            "message": "The IAM role arn:aws:iam::123456789012:role/SomeLambda-codedeploy does not give you permission to perform operations in the following AWS service: AWSLambda. Contact your AWS administrator if you need help. If you are an AWS administrator, you can grant permissions to your users or groups by creating IAM policies."
        },
```

The role is present, and has the correct policies attached (including `AWSCodeDeployForLambda`). 

It turns out the above error is because I'm passing in a hook named `something_like_this`, and the managed policy expects hooks to conform to the ARN: `arn:aws:lambda:*:*:function:CodeDeployHook_*` - it might be nice to document this: for custom hooks, either name them beginning with `CodeDeployHook_`, or pass in custom role.

If we really want to be nice to end users, though, we could do better: if `var.before_allow_traffic_hook_arn` (or the "after" equivalent) is present _and_ `var.create_codedeploy_role` is set, invoke permissions could be included in the policy automatically.

dominics:
Updates the deploy README to mention the naming restrictions on the hook input variables. (The other changes to the aws version noted in the README appear to be pre-existing.)

No breaking changes, just a documentation update.

Fixes #77 - pretty much. I would have updated the `deploy` example, but I don't have a deploy lifecycle hook lambda example for Python (only Typescript), so I left that for now.

### DIFF
--- a/modules/deploy/README.md
+++ b/modules/deploy/README.md
@@ -100,7 +100,7 @@ module "lambda" {
 | Name | Version |
 |------|---------|
 | terraform | >= 0.12.6 |
-| aws | >= 2.67 |
+| aws | >= 3.19 |
 | local | >= 1 |
 | null | >= 2 |
 
@@ -108,7 +108,7 @@ module "lambda" {
 
 | Name | Version |
 |------|---------|
-| aws | >= 2.67 |
+| aws | >= 3.19 |
 | local | >= 1 |
 | null | >= 2 |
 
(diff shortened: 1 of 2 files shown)


antonbabenko:
Thanks @dominics !

[v1.33.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.33.0) has been just released.

PR #85, opened 2020-12-11
fix: skip creating deployments if current and target versions match

### LINKED ISSUE #82: After a successful deploy, next run will cause a failed deploy if target version is not updated
TF: 0.14.2 (but present on 0.13.x too)
Module: 1.31.0

I'm running into an issue where the deploy module fails every _second_ time it deploys. The error is:

> The deployment failed because the AppSpec file that specifies the AWS Lambda deployment configuration is missing or has an invalid configuration.

Normally there's an extra part at the end of this message that says what is wrong. Not in this case. I've double-checked the content of the AppSpec file, and it looks correct.

## TargetVersion must not equal CurrentVersion

I started to look deeper. I've been using this script, which is similar to the guts of the deploy script the module uses, but lets me freely control the target and current version easily, and muck with the create-deployment call.

```bash
from=123
to=123

content="{\"Resources\":[{\"MyFunction\":{\"Properties\":{\"Alias\":\"deployed\",\"CurrentVersion\":\"$from\",\"Name\":\"some_function_name\",\"TargetVersion\":\"$to\"},\"Type\":\"AWS::Lambda::Function\"}}],\"version\":\"0.0\"}"
sum=$(echo -n "$content" | shasum -a 256 | cut -f 1 -d ' ')
escaped="${content//\"/\\\"}"
revision="{\"revisionType\": \"AppSpecContent\", \"appSpecContent\": {\"content\": \"$escaped\", \"sha256\": \"$sum\"}}"

aws deploy --debug create-deployment \
    --application-name SomeAppName \
    --deployment-group-name SomeDeploymentGroupName \
    --revision "$revision"
```

What I found is that any time the `CurrentVersion` and `TargetVersion` are different, the deploy succeeds. But any time the versions are the same, the above error is given. It'd be useful to check that hypothesis more widely, but it definitely holds according to my experimentation above. (**Edit: also needed to reproduce is to ensure the deployment group is not using an all-at-once config**)

~I'm working to get this~ **confirmed** by AWS support

### Consequences

If triggering a deploy when one isn't needed (because the versions already match) always causes a failed deploy, we should not CreateDeployment if TargetVersion == CurrentVersion

It also means that the `force_deploy` option will cause failed deployments unless the target version changes every time we apply.

### Incorrect guesses

I had some earlier speculation that this was caused by an AWS bug at https://github.com/aws/serverless-application-model/issues/291 (more references at https://forums.aws.amazon.com/message.jspa?messageID=846025). On reflection, it doesn't seem to be.

dominics:
It appears AWS CodeDeploy will reject deployments where the CurrentVersion and TargetVersion match (edit: if the deployment config is anything other than all-at-once, e.g. canary or linear rollout). We avoid creating such a failing deployment by ignoring cases where the current and target versions already match in the deploy script.

Fixes #82, where there's a lot more detail on the motivation here.

### DIFF
--- a/modules/deploy/main.tf
+++ b/modules/deploy/main.tf
@@ -34,7 +34,7 @@ locals {
   script = <<EOF
 #!/bin/bash
 
-if [[ "${var.target_version}" == "${var.current_version != "" ? var.current_version : local.current_version}" ]]; then
+if [[ '${var.target_version}' == '${var.current_version != "" ? var.current_version : local.current_version}' ]]; then
   echo "Skipping deployment because target version (${var.target_version}) is already the current version"
   exit 0
 fi
--- a/modules/deploy/main.tf
+++ b/modules/deploy/main.tf
@@ -33,6 +33,12 @@ locals {
 
   script = <<EOF
 #!/bin/bash
+
+if [[ "${var.target_version}" == "${var.current_version != "" ? var.current_version : local.current_version}" ]]; then
+  echo "Skipping deployment because target version (${var.target_version}) is already the current version"
+  exit 0
+fi
+
 ID=$(${var.aws_cli_command} deploy create-deployment \
     --application-name ${local.app_name} \
     --deployment-group-name ${local.deployment_group_name} \


modules/deploy/main.tf
```diff
@@ -33,6 +33,12 @@ locals {
 
   script = <<EOF
 #!/bin/bash
+
+if [[ "${var.target_version}" == "${var.current_version != "" ? var.current_version : local.current_version}" ]]; then
```
dominics, line 37:
```suggestion
if [[ '${var.target_version}' == '${var.current_version != "" ? var.current_version : local.current_version}' ]]; then
```

Otherwise weirds things happen when the version is $LATEST (we don't turn on `set -u`, so that would silently be replaced with "")

antonbabenko:
[v1.34.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.34.0) has been just released.

👍

PR #88, opened 2020-12-22
docs: Improved package.py error message for missing source_paths

ers81239:
## Description
Changed the wording and included the failing `source_path` when it can't be found.  Added explanation to both the error message and the README.md regarding how relative paths are interpreted.

## Motivation and Context
I spent 30 minutes trying to interpret "source_path must be set" when it was set, but its relativity was wrong.

## Breaking Changes
None.

## How Has This Been Tested?
Manual testing with valid and invalid paths.  This is my first time using the module, so I don't have a complex source_path, so complex source_paths aren't tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,7 @@ terraform apply
 
 ## <a name="build"></a> Build Dependencies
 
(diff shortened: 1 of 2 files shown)


antonbabenko:
Thanks for the addition!

PR #90, opened 2020-12-23
docs: Explicitly state the IAM role property used for lambda_role

davidpchardy:
## Description
Minor change to README to indicate the IAM role ARN should be used when using lambda_role

## Motivation and Context
Since the only requirement is a string, you could pass the IAM role name, or other string type properties.
The error on apply (ValidationException) isn't inherently descript enough to know exactly what was wrong.

## Breaking Changes
none

## How Has This Been Tested?
Reviewed Markdown


antonbabenko:
Good point! Could you also change the `description` for the variable inside `variables.tf` since it is the origin of the content of README.md?

davidpchardy:
Added

antonbabenko:
Thank you!

PR #94, opened 2021-01-14
fix: Fixed apigateway trigger to use source_arn

### LINKED ISSUE #78: Resource Policy Trigger doesn't work for API Gateway Custom Authorizers
The code for lambda permissions checks to see if the service == apigateway, and if it does, it adds /*/*/* to the end of the value.  For lambda authorizers, this value needs to end in /authorizers/*.

antonbabenko:
Fixes #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,8 +277,8 @@ module "lambda_function" {
 
   allowed_triggers = {
     APIGatewayAny = {
-      service = "apigateway"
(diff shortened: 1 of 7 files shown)


antonbabenko:
[v1.32.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.32.0) has been just released.

PR #99, opened 2021-01-22
Fix memory size limit

bayazee:
## Description
Update memory size limit.

## Motivation and Context
Max memory size limit is increased from 3008MB to 10GB.


antonbabenko:
Thanks for the update!

PR #100, opened 2021-01-26
fix(iam): add permission to create log group

svenlito:
## Description
add iam CreateLogGroup permission

## Motivation and Context
To be in line with the `AWSLambdaBasicExecutionRole`.

### DIFF
--- a/iam.tf
+++ b/iam.tf
@@ -1,5 +1,5 @@
 locals {
-  create_role = var.create && var.create_function && ! var.create_layer && var.create_role
+  create_role = var.create && var.create_function && !var.create_layer && var.create_role
 
   # Lambda@Edge uses the Cloudwatch region closest to the location where the function is executed
   # The region part of the LogGroup ARN is then replaced with a wildcard (*) so Lambda@Edge is able to log in every region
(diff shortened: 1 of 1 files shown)


iam.tf
```diff
@@ -57,6 +57,7 @@ data "aws_iam_policy_document" "logs" {
     effect = "Allow"
 
     actions = [
+      "logs:CreateLogGroup",
```
antonbabenko, line 60: This one action should be only allowed if `!var.use_existing_cloudwatch_log_group`. Will you be able to update the code?

svenlito: @antonbabenko I'll have a look at it.

iam.tf
```diff
@@ -56,11 +56,10 @@ data "aws_iam_policy_document" "logs" {
   statement {
     effect = "Allow"
 
-    actions = [
-      "logs:CreateLogGroup",
+    actions = concat([
       "logs:CreateLogStream",
       "logs:PutLogEvents",
-    ]
+    ], compact([!var.use_existing_cloudwatch_log_group ? "logs:CreateLogGroup" : ""]))
```
svenlito, line 62: @antonbabenko not sure if this is ok?

antonbabenko: Almost, e9ef635

antonbabenko:
[v1.35.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.35.0) has been just released.

PR #111, opened 2021-02-18
feat: Add output for lambda CloudWatch log group name

### LINKED ISSUE #110: Output log group name
I need to use the log group name on a metric filter, but only the ARN is output, so I will need to break apart the ARN to reliably get the name.

Glen-Moonpig:
## Description

New output for the log group name.

## Motivation and Context

Resolves #110 
I need the log group name for a metric filter resource.

## Breaking Changes

None.

## How Has This Been Tested?

Can see the output value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -685,6 +685,7 @@ Q4: What does this error mean - `"We currently do not support adding policies fo
 | Name | Description |
 |------|-------------|
 | lambda\_cloudwatch\_log\_group\_arn | The ARN of the Cloudwatch Log Group |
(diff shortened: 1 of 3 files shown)


antonbabenko:
Thanks @Glen-Moonpig for this addition!

Normally, we only output attributes of the resources, but in this case, `name` is also helpful.

[v1.38.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.38.0) has been just released.

PR #112, opened 2021-02-21
chore: add ci-cd workflow for pre-commit checks

bryantbiggs:
## Description
- add ci-cd workflow for pre-commit checks

## Motivation and Context
- provide static checks for pull requests

## Breaking Changes
- no

## How Has This Been Tested?
- copy of what is implemented in SQS module repository

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,55 @@
+name: Pre-Commit
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  getBaseVersion:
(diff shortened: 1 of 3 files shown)


antonbabenko:
Couple of issues and ideas - https://github.com/terraform-aws-modules/terraform-aws-lambda/runs/1947301594?check_suite_focus=true :

1. I think we need to verify and panic only if anything fails when using `max version`
2. I think we need to disable `terraform_fmt`, `terraform_docs`, `terraform_tflint` checks when using `min version` because Terraform updates fmt and terraform-docs changing behavior from time to time.

PR #113, opened 2021-02-21
chore: Bump versions in examples for GH Actions to pass

antonbabenko:
Each example may have different version requirements.

@bryantbiggs What do you think we can do as a workaround?

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,55 @@
+name: Pre-Commit
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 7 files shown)


bryantbiggs:
ya, I'm not sure if I can do that and use pre-commit. we can treat each example separately + the module as a whole (meaning grab the min/max from the module and run the checks at the root of the project, and then do the same for each example - this allows us to pick up the different versions if there are any). for now I have added this https://github.com/terraform-aws-modules/terraform-aws-lambda/pull/114 to only run validate on the minimum supported version, but all checks on the maximum supported version

antonbabenko:
Closing in favor of #114

Issue #83, opened 2020-12-10
import no longer supported with Terraform 0.12.x as of v1.29.0


blbradley:
on Terraform 0.12.16 and 0.12.29 with v1.31.0

```
✗ terraform import "module.lambda.aws_lambda_function.this[0]" my-function
module.lambda.aws_lambda_function.this[0]: Importing from ID "my-function"...
module.lambda.aws_lambda_function.this[0]: Import prepared!
  Prepared aws_lambda_function for import
module.lambda.aws_lambda_function.this[0]: Refreshing state... [id=my-function]

Error: Error in function call

  on .terraform/modules/lambda/iam.tf line 9, in locals:
   9:   role_name = local.create_role ? coalesce(var.role_name, var.function_name) : null
    |----------------
    | var.function_name is ""
    | var.role_name is null

Call to function "coalesce" failed: no non-null, non-empty-string arguments.
```

Cleared by switching to v1.28.0, which does not have the line with the error.

antonbabenko:
You need to specify some value for either `function_name` or `role_name` inside of `module` block before calling `terraform import`, like this:

```terraform
module "lambda" {
  source = "terraform-aws-modules/lambda/aws"

  function_name = "my-function"
  handler       = "index.lambda_handler"
  runtime       = "python3.8"
}
```

There is not much we can do in this module, unfortunately!

blbradley:
Function name and role name were already defined.

antonbabenko:
I see the error you have when I don't have `function_name` argument in the module block and it works successfully when I specify it.

bmurphey:
I confirmed I'm getting the same results, with Terraform 0.12.30 and module v1.35.0.  Switching to module v1.28.0 works.

My code:
```
module "main-lambda-trigger" {
  source = "terraform-aws-modules/lambda/aws"
  version = "1.35"

  function_name = "trigger"
  description   = "My awesome lambda function"
  handler       = "trigger.lambda_handler"
  runtime       = "python3.8"

  create_package = false

  tags = {
    Name = "trigger"
  }
}
```

Specifying `role_name` doesn't help, either.

mike-carey:
I have created a PR that could potentially worked around this.  The other option I was able to use was to pin it to a lower version to do the imports and then raise the version back.

antonbabenko:
[...]

PR #117, opened 2021-02-27
fix: revert module Terraform 0.13.x version upgrade 

bryantbiggs:
## Description
- revert module Terraform 0.13.x version upgrade
- update ci-cd checks to validate each Terraform root directory 

## Motivation and Context
- min version used will vary greatly, this update allows examples to have various minimum versions of Terraform and still be tested separate from what the module min version is itself

## Breaking Changes
- No

## How Has This Been Tested?
- CI/CD checks

PR #121, opened 2021-03-02
fix: Defaults the role_name coalesce to * to workaround import error

mike-carey:
## Description

Defaults the role_name to `*` within the coalesce to workaround #83.  By setting a dummy value as the default option, terraform can successfully import without erroring out because the variables are empty and null.  If one still forgot to set both values, an error will be thrown when attempting to create any of the IAM resources with the `*` value.

```
Error: invalid value for name (must match [\w+=,.@-])

  on .terraform/modules/application.lambda/iam.tf line 71, in resource "aws_iam_policy" "logs":
  71:   name   = "${local.role_name}-logs"

```

## Motivation and Context

We still use terraform 0.12 and would like to upgrade to at least v1.29.0 at least up to v1.39.0.  See #83 for details involving the current issue.

## Breaking Changes

No breaking changes.  If anyone is using v1.39.0, it is locked to terraform 0.13 anyways and it seems this bug is specific to 0.12.  However, I wanted to put this out there in anticipation of potentially reverting the minimum version to 0.12.

## How Has This Been Tested?

When using the versions of the module anywhere from `v1.29.0` to `v1.38.0`, I attempted to import resources only to result in the error specified in #83.  When changing the terraform file directly, (editing `.terraform/modules/lambda/iam.tf`), the import then works.  Commenting out the function name and role name being passed in results in the validate error above.

### DIFF
--- a/iam.tf
+++ b/iam.tf
@@ -7,7 +7,11 @@ locals {
   log_group_name         = element(concat(data.aws_cloudwatch_log_group.lambda.*.name, aws_cloudwatch_log_group.lambda.*.name, [""]), 0)
   log_group_arn          = local.create_role && var.lambda_at_edge ? format("arn:%s:%s:%s:%s:%s", data.aws_arn.log_group_arn[0].partition, data.aws_arn.log_group_arn[0].service, "*", data.aws_arn.log_group_arn[0].account, data.aws_arn.log_group_arn[0].resource) : local.log_group_arn_regional
 
-  role_name = local.create_role ? coalesce(var.role_name, var.function_name) : null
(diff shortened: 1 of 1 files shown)


antonbabenko:
Thanks for the PR!

The minimum version of Terraform is reverted to 0.12.26 in #117 and available in [v1.40.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.40.0) of this module, so this PR is relevant for the `master` branch I believe.

antonbabenko:
[v1.43.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.43.0) has been just released.

PR #133, opened 2021-03-25
fix: make lambda function depend on the Cloudwatch log group

Neki:
## Description

Make the AWS Lambda function explicitly depend on the Cloudwatch log group. With this change, Terraform will create the log group before the lambda, and delete it after deleting the lambda.

This avoids a race condition where AWS creates the log group before Terraform can (see below). 

## Motivation and Context

If the lambda function writes logs when the log group doesn't exist, then AWS will create it, and Terraform will error with "log group already exists" on the following `terraform apply`.

It's possible to trigger this case by running `terraform destroy` when the lambda function is being invoked: the log group will be
deleted by Terraform and recreated by AWS before the lambda function is destroyed.

This change imposes an ordering between the lambda function and the log group creation/destruction, avoiding this race condition.

## Breaking Changes

No breaking change, no module API change.

## How Has This Been Tested?

I ran `terraform init && terraform apply && terraform destroy` in the `complete/` example.
Additionally, I've been using this change in production for one month now (with repeated `terraform apply/terraform destroy`), and it fixes the issue described above (that we observed fairly consistently).

### DIFF
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -26,7 +26,7 @@ module "lambda_function" {
   source_path = "${path.module}/../fixtures/python3.8-app1"
 
   store_on_s3 = true
-  s3_bucket   = module.s3_bucket.this_s3_bucket_id
+  s3_bucket   = module.s3_bucket.s3_bucket_id
 
   layers = [
     module.lambda_layer_local.this_lambda_layer_arn,
@@ -183,7 +183,7 @@ module "lambda_layer_s3" {
   source_path = "${path.module}/../fixtures/python3.8-app1"
 
   store_on_s3 = true
-  s3_bucket   = module.s3_bucket.this_s3_bucket_id
+  s3_bucket   = module.s3_bucket.s3_bucket_id
 }
 
 ##############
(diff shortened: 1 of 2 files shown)


Neki:
The CI failure seems unrelated to this change, let me know if I need to add something to this PR :smiley:

svenlito:
@Neki Thanks for your PR. If you could merge in the latest changeset from the master branch, please. I'd help sort out the build issue.

Neki:
Done, the build issue is fixed, thanks.

antonbabenko:
@svenlito thanks for the mention. I will try to take a look at this and other PRs you assign to me at the beginning of the next, after Easter break.

svenlito:
> @svenlito thanks for the mention. I will try to take a look at this and other PRs you assign to me at the beginning of the next, after Easter break.

SGTM, enjoy your break!

antonbabenko:
I missed this PR after I came back from my break. I apologize for that.

[v1.48.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.48.0) has been just released.

PR #141, opened 2021-04-06
fix: permission for lambda-to-lambda async calls

0xe1d1a:
The async policy is missing the action to asynchronously call other lambda functions. Duplicate of #140, which I messed up somehow due to using github UI for the first time.

## Description
Added lambda:InvokeFunction into the policy document for async integrations.

Check https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html as it states

"This operation requires permission for the lambda:InvokeFunction action."

## Motivation and Context
When I deployed a lambda of which the destination_on_success is another lambda arn, my caller lambda failed to call my callee lambda due to a permission error.

## Breaking Changes
Does not break backwards compatibility as it is an added permission.

## How Has This Been Tested?
After adding the above changes on my local terraform module, the lambda integration works as expected, i.e. it has the permission to call another lambda asynchronously.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -94,7 +94,7 @@ jobs:
       - name: Install pre-commit dependencies
         run: |
           pip install pre-commit
-          curl -L "$(curl -s https://api.github.com/repos/terraform-docs/terraform-docs/releases/latest | grep -o -E "https://.+?-v0.12.0-linux-amd64" | head -n1)" > terraform-docs && chmod +x terraform-docs && sudo mv terraform-docs /usr/bin/
+          curl -L "$(curl -s https://api.github.com/repos/terraform-docs/terraform-docs/releases/latest | grep -o -E "https://.+?-v0.12\..+?-linux-amd64" | head -n1)" > terraform-docs && chmod +x terraform-docs && sudo mv terraform-docs /usr/bin/
           curl -L "$(curl -s https://api.github.com/repos/terraform-linters/tflint/releases/latest | grep -o -E "https://.+?_linux_amd64.zip")" > tflint.zip && unzip tflint.zip && rm tflint.zip && sudo mv tflint /usr/bin/
       - name: Execute pre-commit
         # Run all pre-commit checks on max version supported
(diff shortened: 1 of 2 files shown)


0xe1d1a:
@antonbabenko, no idea why this is failing the check but have a look when you can as the error seems like there is something wrong with the runner environment.

antonbabenko:
Thanks for the fix! Looks good to me.

@bryantbiggs Do you have an idea what can this be with GH Actions? `terraform-docs` is giving us extra work lately :(

bryantbiggs:
@antonbabenko ya, I don't have a good solution for `terraform_docs` unless we take it off the integration process and run it manually as part of the release process. there are two big issues that are competing:
1. Terraform docs have been getting a lot of changes lately so we moved to try to pin the version, but this introduces a new issue which is
2. If we pin the version of `terraform_docs`, how do users also utilize that same version. Without something like `tfenv` or `pyenv`, etc. to manage versions, I would suspect most users are pulling the latest or trying to use whatever version is currently installed

something we need to think about how we want to support this, and how it fits into our static checks process

antonbabenko:
I wonder what has changed recently so that we have this error ([failed run](https://github.com/terraform-aws-modules/terraform-aws-lambda/pull/141/checks?check_run_id=2279338377)):

```
Terraform docs...........................................................Failed
- hook id: terraform_docs
- exit code: 1

ERROR: terraform-docs is required by terraform_docs pre-commit hook but is not installed or in the system's PATH.
```

I am fine with running `terraform-docs` myself with the correct version locally in the meantime as long as GH Actions run them also.

bryantbiggs:
`terraform_docs` released v0.12.1 which caused the breakage, you can see a fix here but not sure if that will work longterm - need to find a longterm solution to avoid this cycle of errors https://github.com/terraform-aws-modules/terraform-aws-alb/pull/191/files#diff-ac5eead3f3ce4863c524fff031a87b7aecccb4a0493df087a4e1c704a1505036R97

antonbabenko:
@bryantbiggs Thanks for the help! This one is passing fine now. Let's see when it break next time and we fix it better :)

Merging this one.

antonbabenko:
[v1.45.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.45.0) has been just released.

Issue #138, opened 2021-03-31
`trusted_entities` input is only accepting `Service` Principals


flibustier:
## Is your request related to a new offering from AWS?
I don’t think so

## Is your request related to a problem? Please describe.
We want our trusted entities from our lambda role to look like the following:
```
"Principal": {
  "Service": "lambda.amazonaws.com",
  "AWS": "arn:aws:sts::************:assumed-role/OktaRoleFor*****/******@*****.com"
},
```

Unfortunately, when using the input `trusted_entities` as following: 
```
trusted_entities = ["arn:aws:sts::************:assumed-role/OktaRoleFor*****/******@*****.com"]
```

Terraform tries to perform this :
```
# module.aws_iam_role.lambda[0] will be updated in-place
  ~ resource "aws_iam_role" "lambda" {
      ~ assume_role_policy    = jsonencode(
          ~ {
              ~ Statement = [
                  ~ {
                      ~ Principal = {
                          - AWS     = "arn:aws:sts::************:assumed-role/OktaRoleFor*****/******@*****.com" -> null
                          ~ Service = "lambda.amazonaws.com" -> [
                              + "lambda.amazonaws.com",
                              + "arn:aws:sts::************:assumed-role/OktaRoleFor*****/******@*****.com",
                            ]
                        }
                        # (3 unchanged elements hidden)
                    },
                ]
                # (1 unchanged element hidden)
            }
        )
```

Which will result in a 
`Error Updating IAM Role (****) Assume Role Policy: MalformedPolicyDocument: Invalid principal in policy: "SERVICE":"arn:aws:sts::************:assumed-role/OktaRoleFor*****/******@*****.com"`

## Describe the solution you'd like.
It would be nice to use `trusted_entities` like so:
```
trusted_entities = {
  "AWS" = "arn:aws:sts::************:assumed-role/OktaRoleFor*****/******@*****.com"
}
```

## Describe alternatives you've considered.
We don’t find an alternative at this moment, we would appreciate any direction :) 

## Additional context
Thank you all for your great work on this fabulous module !! 🎉

milosbackonja:
@flibustier You could create role and policy separately, and specify role ARN in lambda together with create_role = false

antonbabenko:
+1 to what @milosbackonja says.

This module does not support the use-case you describe out-of-the-box because it is less common.

PR #143, opened 2021-04-15
feat: Extended `trusted_entities` variable to support multiple types

flibustier:
## Description
This change allows a better customization for the `trusted entities` used in the assume role policy

For example :
```hcl
trusted_entities = [
    {
      type = "AWS",
      identifiers = ["arn:aws:sts::XXXXXXXXXXXX:assumed-role/OktaRoleForXXX/xxxx@xxxx"]
    }
]
```

## Motivation and Context
Currently, `trusted_entities` input variable only accepts a list of strings which is added to the type `Service` of the assume role Principal.

For example:
```hcl
trusted_entities = ["service.amazonaws.com"]
```
Will be interpreted as :
```
Principal = {
   ~ Service = "lambda.amazonaws.com" -> [
      + "lambda.amazonaws.com",
      + "service.amazonaws.com",
    ]
}
```

In the issue https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/138 I explain that we need a `AWS` Principal in our assume role (the goal is to execute lambda in local using its assume role).

## Breaking Changes
This first version overwrites `trusted_entities` input variable. It was previously a list of (service only) strings, and now becomes a list of objects consisting of a `type` string (like `Service` or `AWS`) and an `identifiers` array of strings (like an array of services URL as before or a list of ARNs), which will break previous usages of this variable as a list of strings.

But It could also be moved to another input variable so it doesn’t break the current `trusted_entities` input variable. The only thing that is missing is a good variable name which reflects the openness of the input variable compared to the current `trusted_entities`, which actually seems to be a `trusted_services_entities`.

## How Has This Been Tested?

- [x] I have executed the code in our project and it results in a good terraform plan :
```
module.lambda.aws_iam_role.lambda[0] will be updated in-place
  ~ resource "aws_iam_role" "lambda" {
      ~ assume_role_policy    = jsonencode(
          ~ {
              ~ Statement = [
                  ~ {
                      ~ Principal = {
                          + AWS     = [
                              + "arn:aws:sts::XXXXXXXXXXX:assumed-role/OktaRoleForXXX/xxx@xxx",
                            ]
                            # (1 unchanged element hidden)
                        }
                        # (3 unchanged elements hidden)
                    },
                ]
                # (1 unchanged element hidden)
            }
        )
        # (7 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```
Thank you for your help and your fantastic work on this module 👏

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -436,8 +436,8 @@ variable "attach_policy_statements" {
 }
 
 variable "trusted_entities" {
-  description = "Lambda Function additional trusted entities for assuming roles (trust relationship)"
-  type        = list(any)
+  description = "List of additional trusted entities for assuming Lambda Function role (trust relationship)"
+  type        = any
   default     = []
 }
 
--- a/variables.tf
+++ b/variables.tf
@@ -437,7 +437,7 @@ variable "attach_policy_statements" {
 
 variable "trusted_entities" {
   description = "Lambda Function additional trusted entities for assuming roles (trust relationship)"
-  type        = list(string)
+  type        = list(any)
   default     = []
 }
 
--- a/README.md
+++ b/README.md
@@ -727,7 +727,7 @@ No modules.
 | <a name="input_tags"></a> [tags](#input\_tags) | A map of tags to assign to resources. | `map(string)` | `{}` | no |
 | <a name="input_timeout"></a> [timeout](#input\_timeout) | The amount of time your Lambda Function has to run in seconds. | `number` | `3` | no |
 | <a name="input_tracing_mode"></a> [tracing\_mode](#input\_tracing\_mode) | Tracing mode of the Lambda Function. Valid value can be either PassThrough or Active. | `string` | `null` | no |
-| <a name="input_trusted_entities"></a> [trusted\_entities](#input\_trusted\_entities) | Lambda Function additional trusted entities for assuming roles (trust relationship) | `list(any)` | `[]` | no |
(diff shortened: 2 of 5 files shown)


variables.tf
```diff
@@ -437,8 +437,11 @@ variable "attach_policy_statements" {
 
 variable "trusted_entities" {
   description = "Lambda Function additional trusted entities for assuming roles (trust relationship)"
-  type        = list(string)
-  default     = []
+  type = list(object({
```
antonbabenko, line 440: Could you change this to `list(any)` and use `try()` to make this PR backward compatible and support previous values also.

I think it should be rather straightforward. Let me know if you have questions.

antonbabenko (changes requested):
I like the proposal and it will be even better if you actually update code in `examples/complete` to show this for real.

variables.tf
flibustier: It’s a very good idea, I will try this and keep you informed on this, thank you very much !

flibustier:
I’ve updated the pull request using `try` but I’m not very familiar with it so don’t hesitate to give me your feedback/modifications 
Thank you !

antonbabenko:
Thanks @flibustier !

I have updated the code and examples.

[v1.47.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.47.0) has been just released.

flibustier:
Great job! I successfully tested it in our project!

Thanks so much!! 🎉

PR #147, opened 2021-04-23
feat: Added `cloudwatch_logs_group_name` variable 

ChristophShyper:
## Description

This change will allow to specify custom Log Group, e.g. to group all logs for one project under similar prefix.
Added also missing dependency in Lambda resource for the Log Group, as it's necessary in many cases.

## Motivation and Context

It allows to group logs for the same project.
For example Lambdas under `/my-project/lambda/My-Lambda-Function/` and API GW under `/my-project/api-gw`.

It also fixes possible issues for creating Lambda function and creating Log Group when it's name is being updated for running integration, e.g. with API Gateway. Lambda may be created and executed before Terraform creates Log Group, so Terraform will fail with error when it tries to do so.

## Breaking Changes

No breaking changes, only new functionality.

## How Has This Been Tested?

I've used as local `source` for my Lambda module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -654,6 +654,7 @@ No modules.
 | <a name="input_attach_policy_statements"></a> [attach\_policy\_statements](#input\_attach\_policy\_statements) | Controls whether policy\_statements should be added to IAM role for Lambda Function | `bool` | `false` | no |
 | <a name="input_attach_tracing_policy"></a> [attach\_tracing\_policy](#input\_attach\_tracing\_policy) | Controls whether X-Ray tracing policy should be added to IAM role for Lambda Function | `bool` | `false` | no |
 | <a name="input_build_in_docker"></a> [build\_in\_docker](#input\_build\_in\_docker) | Whether to build dependencies in Docker | `bool` | `false` | no |
+| <a name="input_cloudwatch_logs_group_name"></a> [cloudwatch\_logs\_group\_name](#input\_cloudwatch\_logs\_group\_name) | Name of custom Log Group to use. | `string` | `""` | no |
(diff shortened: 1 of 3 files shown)


antonbabenko:
https://docs.aws.amazon.com/lambda/latest/dg/monitoring-cloudwatchlogs.html :
> Lambda automatically integrates with CloudWatch Logs and pushes all logs from your code to a CloudWatch Logs group associated with a Lambda function, which is named /aws/lambda/<function name>.

If for some reason you want to create a Cloudwatch log outside of the module you can do so by setting `use_existing_cloudwatch_log_group = true`.

I don't see a value in this PR, closing it.

ChristophShyper:
Then how can I use it if it's created outside? What magic will tell Lambda to use mine Log Group?

antonbabenko:
I see you edited your message.

No problems, it happens to all of us :)

Have a great weekend and thanks for using the part(s) of the [serverless.tf](https://serverless.tf) !

ChristophShyper:
It's interesting how human brain works. I was able to fork the repo, create PR, produce a valid use case, even add additional feature, got frustrated about rejecting my use case, but never thought that I only referred Log Group in `depends_on` and no other place and my change is simply stupid.
Today I was modifying numerous modules adding custom Log Groups, my brain just shut down at some point and I did all of that mechanically without even thinking.
Just like driving a car, obeying all traffic rules, paying attention to surroundings, but thinking about something completely else and noticing that I run past destination.

Once again, I'm sorry for taking your time and for being annoyed by your valid response.
I like `serverless.tf` and its very versatile approach to all dependencies in provided modules.

Have a great weekend too!

PR #152, opened 2021-05-20
feat: Added destination_config in aws_lambda_event_source_mapping

byrneo:
## Description
Uncommenting `destination_config` block within event_source_mapping. Git pre-commit hook performed a `terraform fmt` resulting in additional formatting changes to unrelated code.

## How Has This Been Tested?

I'm really not familiar with this module, but I tested it locally by specifying an `event_source_mapping` config similar to what is in `examples/event-source-mapping`. A `plan` and `apply` worked just fine and i could see that the "On-Failure" destination config was now configured under a dynamoDB trigger for my lambda function. I re-ran a `plan` again afterwards just to validate that no other spurious side-effects/changes were detected.

### DIFF
--- a/examples/event-source-mapping/main.tf
+++ b/examples/event-source-mapping/main.tf
@@ -29,12 +29,11 @@ module "lambda_function" {
     dynamodb = {
       event_source_arn  = aws_dynamodb_table.this.stream_arn
       starting_position = "LATEST"
-      # This can be created but it won't be updated/removed. To be reviewed in the future.
-      #      destination_config = {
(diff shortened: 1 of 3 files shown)


antonbabenko:
The reformatting was happening because you use an older version of Terraform (0.13, if I remember correctly).

I have fixed the code a bit to make it similar to the rest.

antonbabenko:
[v2.1.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v2.1.0) has been just released.

PR #157, opened 2021-06-07
docs: Updated README with S3 bucket id handling

zekth:
## Description

Adding precisions regarding the S3 bucket handling.

## Motivation and Context

I faced validation issues because i was using the `arn` instead of the `id` of the bucket. As i saw similar people having this issue on internet, i thought it would be better to clarify this in the documentation.

## Breaking Changes

None

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -518,7 +518,7 @@ variable "s3_object_storage_class" {
 }
 
 variable "s3_bucket" {
-  description = "S3 bucket id to store artifacts"
+  description = "S3 bucket to store artifacts"
   type        = string
   default     = null
 }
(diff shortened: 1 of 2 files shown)


variables.tf
```diff
@@ -518,7 +518,7 @@ variable "s3_object_storage_class" {
 }
 
 variable "s3_bucket" {
-  description = "S3 bucket to store artifacts"
+  description = "S3 bucket id to store artifacts"
```
antonbabenko, line 521:
```suggestion
  description = "S3 bucket to store artifacts"
```

antonbabenko (approved):
LGTM

antonbabenko:
Thanks, @zekth !

[v2.4.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v2.4.0) has been just released.

Issue #160, opened 2021-06-15
Add support for Docker Image creation and upload as a part of Lambda/Layer creation.


Barteus:
## Problem
There is a deployment to the S3 bucket, but there is no easy way to create Docker Image and deploy lambda using this Image

## Proposed solution
It would be nice to be able to name an ECR (own or created in this module), create an Image and create lambda using this Image. There is a nice docker plugin we can leverage to do so - https://registry.terraform.io/providers/kreuzwerker/docker/latest 

## Alternative - use now
The natural alternative is to keep the artefact creation out of infrastructure creation.

antonbabenko:
Thanks for opening this issue.

I agree that it would be helpful to know how to build Docker images and make them used by this module but I also think that it should not be a required part of the module (rather a submodule).

I have recently published a similar behavior in another module - https://github.com/cloudquery/terraform-aws-cloudquery/blob/34c04bd391c24d67c9dc3b026f6dffac96230354/main.tf#L111-L112

If you or anyone wants to implement it here and update documentation, it would be great.

Barteus:
Please assign this task to me. I will add ECR as a part of optional variable - similar way like in cloudquery repo.

rogermyung:
I've started work on adding support for container images

antonbabenko:
@Barteus Good, the main difference between this Lambda module and Cloudquery repo is that the Cloudquery repo is using this module. Let's make all docker building resources for Lambda in a submodule.

Barteus:
@antonbabenko I have created a draft for the implementation (https://github.com/Barteus/terraform-aws-lambda/tree/160-docker-image-support).
I have a problem with dependencies for examples - complete and with-etf.
The error messages on terraform init it:
`
Error: Module does not support for_each

on main.tf line 268, in module "lambda_function_for_each":
 268:   for_each = toset(["dev", "staging", "prod"])

Module "lambda_function_for_each" cannot be used with for_each because it
contains a nested provider configuration for "docker", at
../../modules/build/main.tf:8,10-18.

This module can be made compatible with for_each by changing it to receive all
of its provider configurations from the calling module, by using the
"providers" argument in the calling module block.
`

I have found some information here: https://www.terraform.io/docs/language/modules/develop/providers.html but I could not solve it in any way. Can you help me a bit here?

antonbabenko:
@Barteus Which version of Terraform do you use?

[...]

PR #162, opened 2021-06-21
feat: Add submodule to handle creation of docker images

Barteus:
## Description
Added new submodule which contains docker image creation logic.
The new submodule has a Usage description and is used in the example `container-image`.

## Motivation and Context
Wants to stay DRY for lambdas which require Image packaging. 

## How Has This Been Tested?
- I have tested and validated these changes using provided `examples/*` projects (I have run them to check if change does not break anything)
- I have added submodule usage using `container-image` example


antonbabenko:
Thanks, @Barteus !

[v2.5.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v2.5.0) has been just released.

Fixes #160

PR #173, opened 2021-07-13
feat: Add unique_id output of the lambda role

pglagerweij:
## Description
Added unique_id output of the created lambda role.

## Motivation and Context
We have s3 bucket policies that use the unique role id to restrict access to specific roles. This make sure I can create the lambda role with the module instead of creating the role separately or using a data resource to get the required variable.

## Breaking Changes
Doesn't break anything as far as I am aware.

## How Has This Been Tested?
Applied this using terraform 13.4 and 1.0 and used the output of this feature


antonbabenko:
Thanks, @pglagerweij for the PR!

PR #175, opened 2021-07-15
feat: Add support for separate deployments of infra and code

anoshi:
## Description
Ignores changes to the source code hash of the lambda function code to maintain an idempotent terraform build in cases where the application code (lambda package) deployment is managed by an external pipeline / party.

Because the content of `lifecycle` `ignore_changes` lists cannot be dynamically generated, this update simply sets the `aws_lambda_function`'s optional  argument `source_code_hash` to `null` if the new variable `ignore_source_code_hash` is set to `true`.

## Motivation and Context
Allow terraform to manage infrastructure components and application developers to manage the application code

## Breaking Changes
none

## How Has This Been Tested?
- [x] I have tested this against multiple concurrent calls to the module with and without the `ignore_source_code_hash` variable specified / set to `true`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 <a name="unreleased"></a>
 ## [Unreleased]
 
-
+- feat: Add `ignore_changes_package` variable to allow the lambda function resource to be managed by terraform but have the function code managed externally
 
 <a name="v2.7.0"></a>
 ## [v2.7.0] - 2021-07-08
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 <a name="unreleased"></a>
 ## [Unreleased]
 
-- feat: Add `ignore_changes_package` variable to allow the lambda function resource to be managed by terraform but have the function code managed externally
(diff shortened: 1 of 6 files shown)


asiegman:
This change would be useful for us as well.

vhsantos:
We have tested it and its works us expected.

antonbabenko (commented):
You can deploy any package using `local_existing_package` argument and then update 
it and run `terraform apply` again to redeploy it.

See this:

https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/afbf7bd3cbe4a476bafe07327793fc9dd7b99cf4/examples/complete/main.tf#L276-L277

There is no need for extra work unless I am missing something important.

anoshi:
G'day Anton, love your work. You've contributed a massive amount to Terraform and its uptake. Thank you!

The purpose of this update is for Terraform _not_ to manage the deployed code but only the lambda resource and associated attributes, triggers, etc. This update allows another group (e.g. pure developers) to publish the lambda code package on their own schedule and via a separate pipeline i.e. not through `terraform apply`.

I don't believe the module currently supports this type of infrastructure vs code behaviour?

antonbabenko:
@anoshi 👍  Thanks for the explanation - I now understand the use case! :)

Let me finish this PR a bit and extend it to work with Lambda Layers also.

antonbabenko:
Thanks for the PR!

[v2.10.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v2.10.0) has been just released.

PR #199, opened 2021-09-25
feat: support self managed kafka in event source mapping

kty1965:
## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -219,7 +219,7 @@ resource "aws_lambda_event_source_mapping" "this" {
 
   function_name = aws_lambda_function.this[0].arn
 
-  event_source_arn = each.value.event_source_arn
+  event_source_arn = each.value.self_managed_event_source == null ? each.value.event_source_arn : null
 
   batch_size                         = lookup(each.value, "batch_size", null)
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -219,7 +219,7 @@ resource "aws_lambda_event_source_mapping" "this" {
 
   function_name = aws_lambda_function.this[0].arn
 
-  event_source_arn = each.value.event_source_arn
+  event_source_arn = each.value.self_managed_event_source == null ? each.value.event_source_arn : null
```
bryantbiggs, line 222:
```suggestion
  event_source_arn = lookup(each.value, "event_source_arn", null)
```

main.tf
```diff
@@ -242,6 +242,13 @@ resource "aws_lambda_event_source_mapping" "this" {
     }
   }
 
+  dynamic "self_managed_event_source" {
+    for_each = lookup(each.value, "self_managed_event_source", [])
+    content {
+      endpoints = self_managed_event_source.value["endpoints"]
```
bryantbiggs, line 248:
```suggestion
      endpoints = self_managed_event_source.value.endpoints
```

PR #214, opened 2021-10-09
chore: Added example for pip install in layers

### LINKED ISSUE #87: pip_requirements in source_path is ignored with layers
Deploying a lambda layer with a `pip_requirements = true` in `source_path` do not call `pip install`.

This is because the `plan()` function in `package.py` uses the `runtime` field (used for lambda functions) to determine whether it can actually call `pip install`, but it does not support `compatible_runtime`.

A workaround is to set a Python `runtime` even if we create a layer.

For instance, the following configuration will not call `pip install`:
```hcl
module "layer" {
  source = "terraform-aws-modules/lambda/aws"
  version = "~> 1.31.0"

  create_function = false
  create_layer = true
  layer_name = "example"
  compatible_runtimes = ["python3.8"]

  source_path = [
    {
      path = "${path.root}/external/lambda/layers/example",
      pip_requirements = true
    }
  ]
}
```
But if we specify a `runtime`, it will.

MBarendregt:
## Description

Just added an example on how to enforce pip install on a layer.

## Motivation and Context

as requested by #87 not a fix, but a workaround

Closes #87

### DIFF
--- a/examples/build-package/README.md
+++ b/examples/build-package/README.md
@@ -35,6 +35,7 @@ Note that this example may create resources which cost money. Run `terraform des
 |------|--------|---------|
 | <a name="module_lambda_function_from_package"></a> [lambda\_function\_from\_package](#module\_lambda\_function\_from\_package) | ../../ |  |
 | <a name="module_lambda_layer"></a> [lambda\_layer](#module\_lambda\_layer) | ../../ |  |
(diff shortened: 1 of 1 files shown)


ivantichy:
What a pity that is is not updated also here https://registry.terraform.io/modules/terraform-aws-modules/lambda/aws/latest

antonbabenko:
That's true. The code in this PR is here - https://registry.terraform.io/modules/terraform-aws-modules/lambda/aws/latest/examples/build-package - but you are right, they are not visible in the main README.

There are a lot more examples in the `examples` folder.

PR #225, opened 2021-11-03
fix: Raise failure when CodeDeploy deployment fails

wgiddens:
closes terraform-aws-modules/terraform-aws-lambda/issues/224

When CodeDeploy fails to finish the deployment, Terraform doesn't realize it and keeps processing. This PR makes the Terraform apply fail whenever the code deploy fails.

### DIFF
--- a/modules/deploy/main.tf
+++ b/modules/deploy/main.tf
@@ -63,6 +63,8 @@ while [[ $STATUS == "Created" || $STATUS == "InProgress" || $STATUS == "Pending"
     sleep 5
 done
 
+${var.aws_cli_command} deploy get-deployment --deployment-id $ID
+
 if [[ $STATUS == "Succeeded" ]]; then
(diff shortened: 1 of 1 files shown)


modules/deploy/main.tf
```diff
@@ -67,6 +67,7 @@ if [[ $STATUS == "Succeeded" ]]; then
     echo "Deployment succeeded."
 else
     echo "Deployment failed!"
+    exit 1
```
antonbabenko, line 70: Good point, but maybe `deploy get-deployment` (line 73) should be also executed before exiting to see the error message? For example, move line 73 before the `if` statement on line 66.

What do you think?

wgiddens: Yes, that makes sense. I switched the order for the case where `wait_deployment_completion` is false so the pattern is consistent.

PR #233, opened 2021-11-18
feat: Added support for random sleep delay in deploy submodule

pkleanthous:
## Description
Adding here 2 minor tweaks to the deploy sub-module that I found useful.

## Motivation and Context
1) When I deploy multiple lambdas and use the `wait_deployment_completion` the AWS is throttling the CLI calls and the script fails. Here, I'm introducing the `get_deployment_sleep_timer` variable that it can increase the sleep time. Also the sleep time is a random number between `(0  and 10) + ${get_deployment_sleep_timer}`.

2) I want to separate the infrastructure deployment from the lambda lambda function release via CodeDeploy. The motivation behind this is to capture any issues that introduced by the infrastructure changes.
An example is, you change the permissions for a resource that a lambda function use, the old version immediately louses access and you wait until the traffic switches to the new version that has all the right permissions. So the `run_deployment` now will trigger the bash script otherwise deploy the infra, save the script and run it whenever you want.

## Breaking Changes
In order to run the deployment bash script now you need to set `run_deployment = true`

## How Has This Been Tested?
- Both features are tested in a private project. Please let me know that evidence I should provide here.

### DIFF
--- a/modules/deploy/main.tf
+++ b/modules/deploy/main.tf
@@ -60,7 +60,10 @@ while [[ $STATUS == "Created" || $STATUS == "InProgress" || $STATUS == "Pending"
         --deployment-id $ID \
         --output text \
         --query '[deploymentInfo.status]')
-    sleep 5
+
+    SLEEP_TIME=$(( $RANDOM % 10 ) + ${var.get_deployment_sleep_timer})
+    echo "Sleeping for: $SLEEP_TIME Seconds"
+    sleep $SLEEP_TIME
 done
 
 ${var.aws_cli_command} deploy get-deployment --deployment-id $ID
@@ -106,7 +109,7 @@ resource "local_file" "deploy_script" {
 }
 
 resource "null_resource" "deploy" {
-  count = var.create && var.create_deployment ? 1 : 0
+  count = var.run_deployment ? 1 : 0
 
   triggers = {
     appspec_sha256 = local.appspec_sha256
--- a/modules/deploy/main.tf
+++ b/modules/deploy/main.tf
@@ -61,7 +61,7 @@ while [[ $STATUS == "Created" || $STATUS == "InProgress" || $STATUS == "Pending"
         --output text \
         --query '[deploymentInfo.status]')
 
-    SLEEP_TIME=$(( $RANDOM % 10 ) + ${var.get_deployment_sleep_timer})
+    SLEEP_TIME=$(( $RANDOM % 5 ) + ${var.get_deployment_sleep_timer})
     echo "Sleeping for: $SLEEP_TIME Seconds"
     sleep $SLEEP_TIME
 done
--- a/modules/deploy/main.tf
+++ b/modules/deploy/main.tf
@@ -109,7 +109,7 @@ resource "local_file" "deploy_script" {
 }
 
 resource "null_resource" "deploy" {
-  count = var.run_deployment ? 1 : 0
+  count = var.create && var.create_deployment && var.run_deployment ? 1 : 0
 
   triggers = {
     appspec_sha256 = local.appspec_sha256
--- a/examples/deploy/main.tf
+++ b/examples/deploy/main.tf
@@ -23,13 +23,6 @@ module "lambda_function" {
 
   source_path = "${path.module}/../fixtures/python3.8-app1"
   hash_extra  = "yo1"
-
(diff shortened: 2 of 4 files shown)


modules/deploy/main.tf
```diff
@@ -106,7 +109,7 @@ resource "local_file" "deploy_script" {
 }
 
 resource "null_resource" "deploy" {
-  count = var.create && var.create_deployment ? 1 : 0
+  count = var.run_deployment ? 1 : 0
```
antonbabenko, line 112:
```suggestion
  count = var.create && var.create_deployment && var.run_deployment ? 1 : 0
```

`run_deployment = true` should be used in addition to the rest.

modules/deploy/main.tf
```diff
@@ -60,7 +60,10 @@ while [[ $STATUS == "Created" || $STATUS == "InProgress" || $STATUS == "Pending"
         --deployment-id $ID \
         --output text \
         --query '[deploymentInfo.status]')
-    sleep 5
+
+    SLEEP_TIME=$(( $RANDOM % 10 ) + ${var.get_deployment_sleep_timer})
+    echo "Sleeping for: $SLEEP_TIME Seconds"
+    sleep $SLEEP_TIME
```
antonbabenko, line 66: Are you sure that `aws deploy get-deployment` is being throttled? Sounds strange because it is a read type of operation even if it is executed multiple times at once.

pkleanthous: Hi @antonbabenko,

Thanks for your review.

In a busy day when we make multiple AWS API calls we get something like:
`An error occurred (ThrottlingException) when calling the GetDeployment operation (reached max retries: 4): Rate exceeded`

I think the following link describes the error (even if it's a few years old): https://forums.aws.amazon.com/thread.jspa?messageID=892511

modules/deploy/main.tf
```diff
@@ -106,7 +109,7 @@ resource "local_file" "deploy_script" {
 }
 
 resource "null_resource" "deploy" {
-  count = var.create && var.create_deployment ? 1 : 0
+  count = var.create && var.create_deployment && var.run_deployment ? 1 : 0
```
antonbabenko, line 112: Looks good. Could you also update related examples in the `examples/deploy` directory and verify them?

modules/deploy/main.tf
antonbabenko: Right. Let's drop the `get_deployment_sleep_timer` variable because it is not necessary, but just sleep for 5 to 10 seconds, like this:

```
SLEEP_TIME=$(( $RANDOM % 5 ) + 5)
```

Short and simpler.

antonbabenko (changes requested):

modules/deploy/main.tf
pkleanthous: I prefer to parameterize the sleep timer, as even if you are shifting 50% of the traffic every minute, you don't need to check the progress every ~ 5 seconds.


Is it a deal breaker for you if we keep the `get_deployment_sleep_timer` and set the default to 5?

antonbabenko: It is not a deal-breaker for me but I was not familiar with the use-case you had, so trying to balance between code simplicity and required flexibility. Let's keep it as you proposed. :)

examples/deploy/versions.tf
```diff
@@ -1,8 +1,8 @@
 terraform {
-  required_version = ">= 0.13.1"
+  required_version = ">= 1.0.0"
```
antonbabenko, line 2:
```suggestion
  required_version = ">= 0.13.1"
```

examples/deploy/versions.tf
```diff
@@ -1,8 +1,8 @@
...
 
   required_providers {
-    aws    = ">= 3.19"
+    aws    = ">= 3.66.0"
```
antonbabenko, line 5:
```suggestion
    aws    = ">= 3.19"
```

antonbabenko (changes requested):
Please revert the version requirements since the change introduced in this PR does not require them, and update the docs. It is almost done.

antonbabenko:
Thanks, @pkleanthous !

[v2.27.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v2.27.0) has been just released.

PR #243, opened 2021-12-28
feat: Add ECR Lifecycle Policy Option to docker-build module

1Mill:
## Description
Let users add ECR lifecycle policy rules to automate the cleaning up of unused images generated by the `docker-build` module.

## Motivation and Context
I use this module to build and publish many Docker Images to ECR and I want to be able to easily delete unneeded Docker Images. Since this module is the one building and publishing the Docker Images to ECR it seems best to extend the module to also support the JSON ECR lifecycle policy document. 

## Breaking Changes
This is not a breaking change and is only added non-breaking functionality to the module. 

## How Has This Been Tested?
```hcl
// examples/container-image/main.tf

module "docker_image" {
  source = "../../modules/docker-build"

  create_ecr_repo = true
  ecr_repo        = random_pet.this.id
  image_tag       = "1.0"
  source_path     = "context"
  build_args = {
    FOO = "bar"
  }
  ecr_repo_lifecycle_policy = <<EOF
{
  "rules": [
    {
      "rulePriority": 1,
      "description": "Keep only the last 2 images",
      "selection": {
        "tagStatus": "any",
        "countType": "imageCountMoreThan",
        "countNumber": 2
      },
      "action": {
        "type": "expire"
      }
    }
  ]
}
EOF
}

```

### DIFF
--- a/modules/docker-build/main.tf
+++ b/modules/docker-build/main.tf
@@ -43,7 +43,7 @@ resource "aws_ecr_repository" "this" {
 }
 
 resource "aws_ecr_lifecycle_policy" "this" {
-  count = length(var.ecr_repo_lifecycle_policy) > 0 ? 1 : 0
+  count = var.ecr_repo_lifecycle_policy != null ? 1 : 0
 
   policy     = var.ecr_repo_lifecycle_policy
   repository = local.ecr_repo
--- a/modules/docker-build/main.tf
+++ b/modules/docker-build/main.tf
@@ -41,3 +41,10 @@ resource "aws_ecr_repository" "this" {
 
   tags = var.ecr_repo_tags
 }
+
+resource "aws_ecr_lifecycle_policy" "this" {
+  count = length(var.ecr_repo_lifecycle_policy) > 0 ? 1 : 0
+
+  policy     = var.ecr_repo_lifecycle_policy
+  repository = local.ecr_repo
+}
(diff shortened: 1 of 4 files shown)


modules/docker-build/main.tf
```diff
@@ -41,3 +41,10 @@ resource "aws_ecr_repository" "this" {
 
   tags = var.ecr_repo_tags
 }
+
+resource "aws_ecr_lifecycle_policy" "this" {
+  count = length(var.ecr_repo_lifecycle_policy) > 0 ? 1 : 0
```
antonbabenko, line 46: `var.ecr_repo_lifecycle_policy` is not a list, but a string (should be `null` by default).

```suggestion
  count = var.ecr_repo_lifecycle_policy != null ? 1 : 0
```

modules/docker-build/variables.tf
```diff
@@ -58,3 +58,9 @@ variable "build_args" {
   type        = map(string)
   default     = {}
 }
+
+variable "ecr_repo_lifecycle_policy" {
+  description = "A JSON formatted ECR lifecycle policy to automate the cleaning up of unused images."
+  type        = string
+  default     = ""
```
antonbabenko, line 65:
```suggestion
  default     = null
```

antonbabenko (commented):
Looks pretty good. Please fix docs (`pre-commit run -a`) to make GH checks green.

1Mill:
@antonbabenko, I committed your changes about the default `null` value and updated the README. I gad quite a few issues with getting `pre-commit` working so may want to double check the `pre-commit run -a` command.

antonbabenko:
This PR is included in version 2.29.0 :tada:

PR #244, opened 2021-12-29
feat: Added support for skip_destroy in Lambda Layer Version

absa-rsuarez:
## Description
Added support for new Lambda Layer Version parameter skip_destroy to retain previous layer versions

## Motivation and Context
This allows those using shared layer versions across the organization to prevent the loss of previous in use versions of the layer

## Breaking Changes
There are no breaking changes with this implementation as it follow the logic of the existing parameter by defaulting to false

## How Has This Been Tested?
There are no existing examples beyond the layer generation. This is a supported parameter from the upstream terraform provider aws_lamba_layer_version resource: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_layer_version


antonbabenko:
This PR is included in version 2.30.0 :tada:

absa-rsuarez:
Thank you @antonbabenko !

PR #256, opened 2022-01-21
fix: Updated code style to use `try()`

### LINKED ISSUE #251: Module crashes when create set to false
## Description

If module is used within a larger Terraform stack and create is set to false to disable, then all Terraform plan and apply operations on the parent stack fail with:

Error: Invalid function argument
│
│   on ../../main.tf line 14, in locals:
│   14:   s3_key            = var.s3_existing_package != null ? lookup(var.s3_existing_package, "key", null) : (var.store_on_s3 ? var.s3_prefix != null ? format("%s%s", var.s3_prefix, replace(local.archive_filename, "/^.*//", "")) : replace(local.archive_filename, "/^\\.//", "") : null)
│     ├────────────────
│     │ local.archive_filename is null
│
│ Invalid value for "str" parameter: argument must not be null.

## Versions
 - Terraform: v1.0.11
 - Provider(s): hashicorp/aws v3.39.0
 - Module: v2.28.0

## Reproduction
1. Set create = false
2. terraform plan

### Code Snippet to Reproduce
  create = false

## Expected behavior
Terraform plan or apply should succeed with no op on resources defined by the module.

## Actual behavior
Terraform plan or apply fails with error.

antonbabenko:
Fixed #251 

And updated code style.

### DIFF
--- a/iam.tf
+++ b/iam.tf
@@ -3,8 +3,8 @@ locals {
 
   # Lambda@Edge uses the Cloudwatch region closest to the location where the function is executed
   # The region part of the LogGroup ARN is then replaced with a wildcard (*) so Lambda@Edge is able to log in every region
-  log_group_arn_regional = element(concat(data.aws_cloudwatch_log_group.lambda.*.arn, aws_cloudwatch_log_group.lambda.*.arn, [""]), 0)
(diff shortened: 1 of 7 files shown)


antonbabenko:
This PR is included in version 2.33.1 :tada:

Issue #281, opened 2022-03-17
Lambda Alias no_refresh and with_refresh are backwards


ajmac3242:
# Description
Lambda Alias `no_refresh` and `with_refresh` are backwards. Currently (for the last two years), the `ignore_changes` lifecycle block has been on the `with_refresh` resource. This causes the alias NOT to refresh to the passed in lambda version when you pass in `refresh_alias = true`.

# Version
- Terraform: 1.1.7
- Module: [Lambda Alias](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/master/modules/alias/main.tf#L48)

# Reproduction
1. Create a lambda function and alias with `with_refresh` set to `true`.
2. Run terraform plan and apply
3. Update something on the lambda function and publish a new version.
4. Run terraform plan and apply
5. Alias will not be updated to the latest version of the lambda function

# Expected Behavior
When `refresh_alias = true` I would expect the alias to be pointed at `var.function_version` that was passed into the module.
When `refresh_alias = false` I would expect the alias to stay pinned to the original `var.function_version` that was passed into the module.

# Actual Behavior
When `refresh_alias = true` the alias stays pinned to the original `var.function_version` that was passed into the module.
When `refresh_alias = false` the alias points at `var.function_version` that was passed into the module.

# Code Change

Move lines of code on lines [48-50](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/master/modules/alias/main.tf#L48) to below line [28](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/master/modules/alias/main.tf#L28)

```
resource "aws_lambda_alias" "no_refresh" {
  count = var.create && !var.use_existing_alias && !var.refresh_alias ? 1 : 0

  name        = var.name
  description = var.description

  function_name    = var.function_name
  function_version = var.function_version != "" ? var.function_version : "$LATEST"

  # $LATEST is not supported for an alias pointing to more than 1 version
  dynamic "routing_config" {
    for_each = length(keys(var.routing_additional_version_weights)) == 0 ? [] : [true]
    content {
      additional_version_weights = var.routing_additional_version_weights
    }
  }

  lifecycle {
    ignore_changes = [function_version]
  }
}

resource "aws_lambda_alias" "with_refresh" {
  count = var.create && !var.use_existing_alias && var.refresh_alias ? 1 : 0

  name        = var.name
  description = var.description

  function_name    = var.function_name
[...]

PR #282, opened 2022-03-17
fix: Alias Module: Move lifecycle hook that ignores lambda_version from wi…

ajmac3242:
## Description
Fixes issue described in #281 

## Motivation and Context
This change will all `with_refresh` to behave as expected.

## Breaking Changes
This will cause anyone who have coded around this issue (uses `no_refresh` instead of `with_refresh`) to no longer have their alias refreshed with the passed in version

## How Has This Been Tested?
-- https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/master/examples/alias/main.tf


ajmac3242:
remove stale label

PR #314, opened 2022-05-23
fix: Removed docker provider block from docker-build submodule

### LINKED ISSUE #267: docker-build declares its own provider and therefore cannot be easily deleted
## Description

The docker-build submodule declares its own docker provider. This means that if I create an instance of this module and later try to delete it, `terraform plan` will fail with an error like this:

```
│ Error: Provider configuration not present
│
│ To work with module.docker_image.docker_registry_image.this (orphan) its original provider configuration at
│ module.docker_image.provider["registry.terraform.io/kreuzwerker/docker"] is required, but it has been removed. This occurs when a provider configuration is removed while objects
│ created by that provider still exist in the state. Re-add the provider configuration to destroy module.docker_image.docker_registry_image.this (orphan), after which you can
│ remove the provider configuration again.
```

This doesn't play well with CI pipelines as the only way to fix this is to manually run `terraform state rm`. Can you instead tell users to properly declare their docker provider in their root module instead?

See https://stackoverflow.com/a/58403262 for more information.

antonbabenko:
Fixed #267

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/antonbabenko/pre-commit-terraform
-    rev: v1.69.0
+    rev: v1.71.0
     hooks:
       - id: terraform_fmt
       - id: terraform_validate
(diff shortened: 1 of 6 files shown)


antonbabenko:
This PR is included in version 3.2.1 :tada:

Issue #317, opened 2022-06-09
Lambda Event Source Mapping remain disabled after re-creating lambdas


hussainbani:
After recreating lambdas. event source mapping remains disabled until terraform is re-applied. The fix should be simple as enabling it by default.

https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/e7bfddf314bf915bfb4b2d58b6740f1d6c376be3/main.tf#L243

This should be **true** by default as the creation is already controlled via for_each.

antonbabenko:
I think you are right, please submit a PR.

antonbabenko:
Fixed in #321

PR #321, opened 2022-06-17
fix: Fixed enabled attribute in Lambda Event Source Mapping by default

hussainbani:
Enabling event source mapping by default.

https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/317#issue-1266390128


antonbabenko:
This PR is included in version 3.3.1 :tada:

PR #325, opened 2022-07-08
feat: Add SAM Metadata resources to enable the integration with SAM CLI tool

moelasmar:
## Description
This change is to add new resources (sam_metadata_*) of type (null_resource) that will maintain some information about the lambda resources exist in the terraform application to enable the integration with SAM CLI.  SAM CLI tool is looking for the lambda resources source code, and the resources that contain the building logic of these resources. SAM CLI can use this information to extend the SAM CLI testing capabilities to the terraform applications.

## Motivation and Context
This change will enable the customers who use Serverless TF module to define Lambda resources to integrate with SAM CLI tool, and so the customers can test their applications either locally or on cloud.

## Breaking Changes
no breaking changes.

## How Has This Been Tested?
As the new resource is just to collect metadata for SAM CLI tool, so it does not require any changes from the modules users side, and so no change in the examples. So I tested that I can run `terraform apply` for the available examples without any issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -655,6 +655,8 @@ No modules.
 | [aws_s3_object.lambda_package](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_object) | resource |
 | [local_file.archive_plan](https://registry.terraform.io/providers/hashicorp/local/latest/docs/resources/file) | resource |
 | [null_resource.archive](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
+| [null_resource.sam_metadata_aws_lambda_function_this](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
+| [null_resource.sam_metadata_aws_lambda_layer_version_this](https://registry.terraform.io/providers/hashicorp/null/latest/docs/resources/resource) | resource |
 | [aws_arn.log_group_arn](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/arn) | data source |
 | [aws_cloudwatch_log_group.lambda](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/cloudwatch_log_group) | data source |
 | [aws_iam_policy.tracing](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy) | data source |
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ This Terraform module is the part of [serverless.tf framework](https://github.co
 3. Create, update, and publish AWS Lambda Function and Lambda Layer - [see usage](#usage).
 4. Create static and dynamic aliases for AWS Lambda Function - [see usage](#usage), see [modules/alias](https://github.com/terraform-aws-modules/terraform-aws-lambda/tree/master/modules/alias).
 5. Do complex deployments (eg, rolling, canary, rollbacks, triggers) - [read more](#deployment), see [modules/deploy](https://github.com/terraform-aws-modules/terraform-aws-lambda/tree/master/modules/deploy).
+6. Use AWS SAM CLI to test Lambda Function - [read more](#sam_cli_integration).
 
 ## Features
 
--- a/README.md
+++ b/README.md
@@ -531,7 +531,7 @@ module "lambda_function_existing_package_from_remote_url" {
 applications. Currently, SAM CLI tool only supports CFN applications, but SAM CLI team is working on a feature to extend the testing capabilities to support terraform applications (check this [Github issue](https://github.com/aws/aws-sam-cli/issues/3154) 
 to be updated about the incoming releases, and features included in each release for the Terraform support feature).
 
-SAM CLI provides two ways of testing, local testing, and testing on-cloud (Accelerate).
+SAM CLI provides two ways of testing: local testing and testing on-cloud (Accelerate).
 
 ### Local Testing
 Using SAM CLI, you can invoke the lambda functions defined in the terraform application locally using the [sam local invoke](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-local-invoke.html)
--- a/README.md
+++ b/README.md
@@ -526,6 +526,35 @@ module "lambda_function_existing_package_from_remote_url" {
 }
 ```
 
+## <a name="sam_cli_integration"></a> How to use AWS SAM CLI to test Lambda Function?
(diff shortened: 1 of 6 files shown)


moelasmar:
@bryantbiggs .. Could you please check this PR.

main.tf
```diff
@@ -108,6 +108,33 @@ resource "aws_lambda_function" "this" {
...
+resource "null_resource" "sam_metadata_aws_lambda_function_this" {
+  count = local.create && var.create_package && var.create_function && !var.create_layer ? 1 : 0
+  triggers = {
+    # This is a way to let SAM CLI correlates between the Lambda function resource, and this metadata
+    # resource
+    resource_name = "aws_lambda_function.this[0]",
+    resource_type = "ZIP_LAMBDA_FUNCTION",
+
+    # The Lambda function source code.
+    original_source_code = jsonencode(var.source_path),
```
antonbabenko, line 123: source_path can be rather complicated type (see examples or readme), I am not sure this will be well supported by SAM. Could you please verify this?

antonbabenko (changes requested):
Looks pretty good. I am excited to see an additional integration with SAM, thank you! It would be great to have at least a paragraph saying about this feature and commands to use this with SAM in readme or in example. (If there is no notion about something in terraform-aws-modules - feature doesn't exist)

main.tf
moelasmar: In Sam Cli, we expect the value of this property to be either String or JSON object. In case if it is a JSON object as we have here in serverless TF module, SAM CLI will expect another property `source_code_property` to let SAM CLI know, where it can find the Lambda resource code path. In Sam Cli , we do not support multiple paths, so in case if the value of this attribute is a list (of strings, or objects), SAM CLI will process only the first item in the list.

moelasmar:
Thanks @antonbabenko for your quick response. Regarding the Readme update, is it ok to add a new section regarding SAM CLI integration before the [FAQ section](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/39e45b65e2c421f5932e5456c3ab4897dc0ee0da/README.md#faq) in this Readme file

bryantbiggs:
@moelasmar I would put it under the [`How to deploy and manage Lambda functions`](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/39e45b65e2c421f5932e5456c3ab4897dc0ee0da/README.md#-how-to-deploy-and-manage-lambda-functions)

moelasmar:
> @moelasmar I would put it under the [`How to deploy and manage Lambda functions`](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/master/README.md#-how-to-deploy-and-manage-lambda-functions)

I founded that it is better to have a separate section for SAM CLI integration, as we plan to support testing only for now, and no plan to support package and deploy in the current project.

@antonbabenko could you please have a look, and please let me know if you have any comments.

main.tf
```diff
@@ -108,6 +108,33 @@ resource "aws_lambda_function" "this" {
...
 
+# This resource contains the extra information required by SAM CLI to provide the testing capabilities
+# to the TF application. The required data is where SAM CLI can find the Lambda function source code
+# and what are the resources that contain the building logic.
+resource "null_resource" "sam_metadata_aws_lambda_function_this" {
+  count = local.create && var.create_package && var.create_function && !var.create_layer ? 1 : 0
+  triggers = {
+    # This is a way to let SAM CLI correlates between the Lambda function resource, and this metadata
+    # resource
+    resource_name = "aws_lambda_function.this[0]",
```
antonbabenko, line 119: Please remove comas after each value on the map.

antonbabenko (commented):
I looked at it in details after I had read the specs. In general, this PR looks pretty good, but I was not able to test it for real without building SAM with [your changes](https://github.com/aws/aws-sam-cli/tree/terraform-hooks).

There is one hard part I want to highlight:
1. Instead of doing the whole cycle for AWS resource extraction (`terraform init`, plan into a file, show as jіon, jq, python, etc), I recommend that this module optionally [creates a local file](https://registry.terraform.io/providers/hashicorp/local/latest/docs/resources/file) in the known location (if the user specifies like `create_sam_metadata = true`), SAM just reads the captured JSON object and Prepare Hook will generate CFN Template (or even Terraform). This will be much faster, more flexible, and transparent for the user.

Maybe I don't know the whole picture, so please correct me (you have my email, too). I will be glad to cooperate with you on this (I am back from vacation, finally)!

moelasmar:
Thanks @antonbabenko for your comment. Let me think more about it, also we can discuss it more in  the meeting we have next week.

README.md
```diff
@@ -526,6 +526,35 @@ module "lambda_function_existing_package_from_remote_url" {
 }
 ```
 
+## <a name="sam_cli_integration"></a> How to use AWS SAM CLI to test Lambda Function?
+[AWS SAM CLI](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-command-reference.html) is an open source tool that help the developers to initiate, build, test, and deploy serverless 
+applications. Currently, SAM CLI tool only supports CFN applications, but SAM CLI team is working on a feature to extend the testing capabilities to support terraform applications (check this [Github issue](https://github.com/aws/aws-sam-cli/issues/3154) 
+to be updated about the incoming releases, and features included in each release for the Terraform support feature).
+
+SAM CLI provides two ways of testing, local testing, and testing on-cloud (Accelerate).
```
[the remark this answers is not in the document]
moelasmar: done

main.tf
moelasmar: done

antonbabenko:
Hi @moelasmar !

Thanks for your time and yesterday's discussion about this PR and SAM/TF integration.

If you don't mind, I'd like to keep this PR open until I can run SAM CLI locally myself and verify that it works as expected. It also doesn't look good for the users if we include SAM CLI commands in the docs, but users can't run them because SAM CLI doesn't have them available (not even in `sam-nightly`).

modules/docker-build/main.tf
```diff
@@ -41,3 +41,19 @@ resource "aws_ecr_lifecycle_policy" "this" {
   policy     = var.ecr_repo_lifecycle_policy
   repository = local.ecr_repo
 }
+
+# This resource contains the extra information required by SAM CLI to provide the testing capabilities
+# to the TF application. This resource will maintain the metadata information about the image type lambda
+# functions. It will contain the information required to build the docker image locally.
+resource "null_resource" "sam_metadata_docker_registry_image_this" {
```
antonbabenko, line 48: Can we drop `_this` from the name in resources?

antonbabenko (commented):
LGTM

antonbabenko (approved):
Looks good!

antonbabenko:
Great work, @moelasmar !

antonbabenko:
This PR is included in version 4.6.0 :tada:

moelasmar:
Thanks @antonbabenko for your support

Issue #323, opened 2022-06-29
Multiple Users Cause lambda.null_resource.archive[0] timestamp Differences


gugaiz:
## Description
I am experiencing the exact same issue explained [here](https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/63).

The solution provided was to upgrade to [v1.26.0](https://github.com/terraform-aws-modules/terraform-aws-lambda/releases/v1.26.0) but I am using v3.3.1, is this a regression?

deybhayden:
I'm seeing this same behavior.

- Module Version: `3.3.1`
- Terraform Version: `v1.2.3`

KajiMaster:
Same too; was hoping I was doing something wrong.

Module Version: 3.3.1
Terrafomr Version v1.2.2

KajiMaster:
I was able to remedy this with the solution in https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/51#issuecomment-681978205
Basically recreate_missing_package = false in the module puts a warning message in the timestamp; effectively disabling time as a measure to recreate. Doesn't prevent new packages from being build like it claims. Needs better wording as a solution for this issue related to 'timestamp'

antonbabenko:
> @KajiMaster said: "Doesn't prevent new packages from being build like it claims."

I just double-checked it again and I see the same expected behavior as before. After the package has been created and `recreate_missing_package = false` is specified, there won't be new zip-files created inside of `builds` directory on sequential `terraform apply` runs.

I am not sure how flexible timestamp tracking can be. Currently, it is either `Yes` (content will be checked based on timestamp, and the package will be recreated, if missing) or `No` (no timestamp will be compared).

Will it help, if a user can choose to track drift using an argument like `hash` (see #331) instead of `timestamp` like now?

savitafulmali:
I have added recreate_missing_package = false still terraform plan shows changes.
Is it work only for terraform apply command.?

PR #331, opened 2022-07-25
fix: Make the archive resource change when its content changes

mhmd3bdo:
## Description

The suggested change is to detect changes in `null_resource`  archive by hash and instead of timestamps.

## Motivation and Context

The `terraform plan` command will always show changes because the `null_resource`  archive is triggered by timestamps which makes TF runs not idempotent.

## Breaking Changes
No

## How Has This Been Tested?

### DIFF
--- a/package.tf
+++ b/package.tf
@@ -59,7 +59,7 @@ resource "null_resource" "archive" {
 
   triggers = {
     filename  = data.external.archive_prepare[0].result.filename
-    timestamp = data.external.archive_prepare[0].result.timestamp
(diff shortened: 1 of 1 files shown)


SubhraB:
What happened to the PR?

mhmd3bdo:
https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/323#issuecomment-1234144646
@antonbabenko I'm still leaning towards tracking the file changes using hash, it's better than the current implementation which uses timestamps, hashes will be always the best way to identify if there are changes in happened in a specific file or not.

PR #337, opened 2022-08-07
feat!: Updated AWS provider to v4, added ECR repo force_delete argument in docker-build module

1Mill:
## Description

Maintain behavior to force delete created `aws_ecr_repository` even if it contains images. 

## Motivation and Context

[AWS v4.22.0](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md#4220-july--8-2022) introduced a breaking change to the behavior of this module. 

Previously, `aws_ecr_repository` resources would always be destroyed, even if they contained images. Now, `aws_ecr_repository` resources are only destroyed if either (a) the repository is completely empty of images or (b) `force_delete` is set to `true`. 

This PR sets `force_delete` to `true` so that any created `aws_ecr_repository` resource is always successfully destroyed on `terraform destroy` to maintain the existing behavior of this module. 

## Breaking Changes

* Requires `aws >= 4.22.0`

## How Has This Been Tested?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,9 @@
+# Environmental secrets
+.env
+
+# Docker
+docker-compose.yml
+
 # Local .terraform directories
 **/.terraform/*
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,3 @@
-# Environmental secrets
-.env
-
-# Docker
-docker-compose.yml
-
 # Local .terraform directories
 **/.terraform/*
 
--- a/modules/docker-build/main.tf
+++ b/modules/docker-build/main.tf
@@ -24,7 +24,7 @@ resource "docker_registry_image" "this" {
 resource "aws_ecr_repository" "this" {
   count = var.create_ecr_repo ? 1 : 0
 
-  force_delete         = var.ecr_force_delete
+  force_delete         = true
   name                 = var.ecr_repo
   image_tag_mutability = var.image_tag_mutability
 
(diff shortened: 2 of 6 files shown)


modules/docker-build/main.tf
```diff
@@ -24,6 +24,7 @@ resource "docker_registry_image" "this" {
 resource "aws_ecr_repository" "this" {
   count = var.create_ecr_repo ? 1 : 0
 
+  force_delete         = true
```
antonbabenko, line 27: Please add `force_delete` variable with default value `true`.

modules/docker-build/versions.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 3.35"
+      version = ">= 4.22.0"
```
antonbabenko, line 7:
```suggestion
      version = ">= 4.22"
```

.gitignore
```diff
@@ -1,3 +1,9 @@
+# Environmental secrets
```
antonbabenko, line 1: Please revert these changes.

antonbabenko (changes requested):

modules/docker-build/main.tf
1Mill: Added `var.ecr_force_delete` with a default value of `true`

.gitignore
1Mill: Removed any changes to `.gitignore`

antonbabenko (approved):

antonbabenko:
Hi @1Mill ! Thank you for this addition!

antonbabenko:
This PR is included in version 4.0.0 :tada:

Issue #219, opened 2021-10-22
Question/Problem with  `docker_image` input


JCapriotti:
## Description

I'm wondering why specifying `docker_image` inconsistently requires either `docker_build_root` or `docker_file` to be provided. It seems that `docker_image` works without the other inputs only if the image is already stored locally.

## Versions
- Terraform: 
  ```
  Terraform v1.0.9
  on darwin_amd64
  ```
- Providers:
  ```
  + provider registry.terraform.io/cyrilgdn/postgresql v1.14.0
  + provider registry.terraform.io/hashicorp/aws v3.61.0
  + provider registry.terraform.io/hashicorp/external v2.1.0
  + provider registry.terraform.io/hashicorp/http v2.1.0
  + provider registry.terraform.io/hashicorp/local v2.1.0
  + provider registry.terraform.io/hashicorp/null v3.1.0
  + provider registry.terraform.io/hashicorp/random v3.1.0
  ```
- Module:
  ```
  2.22.0
  ```

## Reproduction

Using these Docker-related inputs (no other docker inputs) AND if you have not pulled the Docker image locally.

```
build_in_docker   = true
docker_image      = "public.ecr.aws/sam/build-python3.9:latest"
```

## Expected behavior

The Docker image is used for the packaging without error.

## Actual behavior

Error:
```
ValueError: docker_build_root or docker_file must be provided
```

### Terminal Output Screenshot(s)
![image](https://user-images.githubusercontent.com/801182/138504675-89345f6d-50d4-414a-b0a6-2c1ba9fee500.png)

## Additional context

This seems to be caused by [a call to `docker images <tag>`](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/c2c4b0484bea441a9625751c5b05ec17962edb96/package.py#L849-L856). 

If the image is found, `ok` is set to `True` and `docker_build_command` is (correctly) never called, which is the method that is raising the error.

This seems like a bug, but I'm not sure of the intent for that block of code.

bellis-ai:
This is still a valid problem. Why the image needs to be pre-pulled is beyond me.

bellis-ai:
If anyone else is encountering this problem and has something like a pipeline where they don't want to add download steps on every apply, use a barebones dockerfile that is "FROM" the image you want

i.e.
```
FROM "public.ecr.aws/sam/build-python3.9" 
```

PR #352, opened 2022-10-12
feat: Add support building layers in DockerInDocker scenario

mwiede:
## Description

We have Jenkins instances running with Docker-In-Docker and there are Pipelines using [Docker images as agent](https://www.jenkins.io/doc/book/pipeline/docker/#workspace-synchronization). I was facing problems with the docker run commands inside of this package, because the volumes for temporary directories or the current workspace/build-root could not be mounted.

The error, which occured was: `ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'`
## Motivation and Context

Give option to make building lambda layer in Docker work in Docker-In-Docker/Jenkins-Docker. To be able to do this, one need to pass the container-id of the parent container to any docker run command in the `--volumes-from` parameter.
On Jenkins this is now possible by defining `TF_VAR_docker_volume` as environent variable.

Additionally, this PR is fixing #219 by pulling the build image if needed.
## Breaking Changes

none.
## How Has This Been Tested?


mwiede:
@antonbabenko what is needed for the PR to be considered?

antonbabenko:
@mwiede I'm at AWS reinvent this and next week, so I can't take a look at this.

Usually, when there is a new feature added we need to update example code to show how to use this feature for real. Could you or someone else make necessary changes there? It would help a lot to speed things up.

PR #354, opened 2022-10-14
feat: add policy_name variable

alisson276:
## Description

Allow to give to the policy a name other than the `var.role_name`

## Motivation and Context

Sometimes we want to import things already created by the AWS Console, and the default policy name is AWSLambdaBasicExecutionRole-<random_uuid>. So in order to be able to import without recreating the policy this change allow to set a new policy name. The default behavior if `var.policy_name` is not set is to use role_name exactly how this was doing before this change.

## Breaking Changes

I don't think so

## How Has This Been Tested?


antonbabenko (approved):

antonbabenko:
This PR is included in version 4.4.0 :tada:

monospacesoftware:
We're experiencing an error with version 4.4.0, appears to be caused by this change:
```
Error: Error in function call

 on .terraform/modules/lambda_package/iam.tf line 15, in locals:
 15:   policy_name = coalesce(var.policy_name, local.role_name)
   ├────────────────
   │ while calling coalesce(vals...)
   │ local.role_name is null
   │ var.policy_name is null

Call to function "coalesce" failed: no non-null, non-empty-string
arguments.
```

antonbabenko:
2 mins, I will correct it.

PR #368, opened 2022-10-27
fix: Qualifiers in event invoke config

sindrig:
## Description
create_current_version_async_event_config and create_unqualified_alias_async_event_config were not working due to zipmap returning a map with qualifier as key and null as value if variables were set to false

## Motivation and Context
there's currently no way to opt out of either current_version or unqualified_version for async event config

## How Has This Been Tested?

I deployed the `examples/alias` example and modified the variables before my change, resulting in no change in terraform state. Then I applied my change and ran apply, and terraform wanted to remove the resources, as it should.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -193,7 +193,7 @@ locals {
 }
 
 resource "aws_lambda_function_event_invoke_config" "this" {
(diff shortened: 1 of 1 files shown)


antonbabenko:
This PR is included in version 4.2.1 :tada:

PR #372, opened 2022-11-02
fix: Opposite refresh_alias behavior. Close #281

### LINKED ISSUE #281: Lambda Alias no_refresh and with_refresh are backwards

micksatana:
## Description
`ignore_changes = [function_version]` lifecycle should be in the `no_refresh` resource.

## Motivation and Context
Its behavior is currently opposite. `refresh_alias = true` should refresh new version to the alias. 

## Breaking Changes
If anyone uses opposite `refresh_alias = false` as a workaround to refresh version, one needs to update to `true` and vice versa.

## How Has This Been Tested?
   - Note: existing Python example, not working in case a developer use `pyenv` so I update to Node.js example instead
   - `provisioned_concurrent_executions` in the example was `1` after run the second `apply` it will failed not allowed to be lower than `10`. This is unrelated to `alias` example. So removed from the example.
Screenshot 2565-11-02 at 23 59 24

   - `terraform` `init`, `plan` then `apply`
   - Update `../fixtures/nodejs14.x-app1/index.js`
   - `terraform apply`
Screenshot 2565-11-02 at 23 50 58
Screenshot 2565-11-02 at 23 54 22


examples/alias/main.tf
```diff
@@ -25,17 +25,15 @@ module "lambda_function" {
 
   function_name = "${random_pet.this.id}-lambda"
   handler       = "index.lambda_handler"
-  runtime       = "python3.8"
+  runtime       = "nodejs14.x"
```
antonbabenko, line 28: Why do you change `runtime` and `source_path`? Is it required for the PR to work? I don't think so.

antonbabenko (changes requested):
Good catch, but please revert unnecessary changes in the example.

examples/alias/main.tf
micksatana: The `python3.8` example doesn't work in case a developer use `pyenv` as the screenshot in the description so I changed to `nodejs14.x` example instead so when new developer contributes he/she won't face the same issue. I can revert back if you prefer.

micksatana: Reverted.

antonbabenko:
This PR is included in version 4.7.1 :tada:

ryandeivert:
hi all -- while I greatly appreciate this fix (it's confused me for a long time), I'm concern that it has semi-silently landed in a patch fix. IMO, this is a **breaking change**, and even using the least egregious upgrade pattern (`~>` [syntax](https://developer.hashicorp.com/terraform/language/expressions/version-constraints#-3)) here would even adopted this breaking change.

I'm curious if there is a way to limit releasing breaking changes to (at a minimum) the minor version releases, even when it is a known-bug fix. unfortunately, this caused numerous functions to unexpectedly stop publishing, and I bet it is also affecting many many others without them noticing 😞

jguttman94:
agreed! I suspect that this issue is a result of this change: https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/383

PR #376, opened 2022-11-11
feat: Added static/defined/computed ARN for the Lambda Function outputs

antonbabenko:


outputs.tf
```diff
@@ -4,6 +4,11 @@ output "lambda_function_arn" {
   value       = try(aws_lambda_function.this[0].arn, "")
 }
 
+output "lambda_function_arn_static" {
+  description = "The static ARN of the Lambda Function. Use this to avoid cycle errors between resources (e.g., Step Functions)"
+  value       = local.create && var.create_function && !var.create_layer ? "arn:aws:lambda:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:function:${var.function_name}" : ""
```
antonbabenko, line 9: @bryantbiggs I rather like this workaround for the situations when we need to create something in advance using `-target=...` before the rest of the resources can use such value. I think this approach is suitable for other cases, too.

antonbabenko:
This PR is included in version 4.7.0 :tada:

outputs.tf
bryantbiggs: Definitely!

PR #389, opened 2022-12-07
feat: Add snap_start functionality

darrenswhite:
## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -102,6 +102,14 @@ resource "aws_lambda_function" "this" {
     }
   }
 
+  dynamic "snap_start" {
+    for_each = var.snap_start_apply_on == null ? [] : [true]
+
+    content {
+      apply_on = var.snap_start_apply_on
+    }
+  }
+
   tags = var.tags
 
   depends_on = [
--- a/variables.tf
+++ b/variables.tf
@@ -217,6 +217,12 @@ variable "image_config_working_directory" {
   default     = null
 }
 
+variable "snap_start_apply_on" {
+  description = "Conditions where snap start is enabled"
+  type        = string
+  default     = null
+}
+
 ###############
 # Function URL
 ###############
--- a/versions.tf
+++ b/versions.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 4.9"
+      version = ">= 4.44"
     }
     external = {
       source  = "hashicorp/external"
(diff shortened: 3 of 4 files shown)


versions.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
```
mmichna, line 4: maybe is better upgrade also terraform version: `required_version = ">= 1.3"`

variables.tf
```diff
@@ -217,6 +217,12 @@ variable "image_config_working_directory" {
   default     = null
 }
 
+variable "snap_start_apply_on" {
```
mmichna, line 220: maybe is better change it to:
```
variable "snap_start" {
  description = "(Optional) Snap start settings for low-latency startups"
  type        = bool
  default     = false
}
```

main.tf
```diff
@@ -102,6 +102,14 @@ resource "aws_lambda_function" "this" {
     }
   }
 
+  dynamic "snap_start" {
```
mmichna, line 105: there missing publish field. Maybe is better use this:
```
  dynamic snap_start {
    for_each = var.snap_start ? [true] : []
    content {
      apply_on = "PublishedVersions"
    }
  }
  publish = var.snap_start
```

faxx1080:
Hi! I wanted to confirm, does this include publishing versions? I have heard from AWS that versions are required for full use of Snap Start.

See: https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html

> You can use SnapStart only on [published function versions](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html#configuration-versions-config) and [aliases](https://docs.aws.amazon.com/lambda/latest/dg/configuration-aliases.html) that point to versions. You can't use SnapStart on a function's unpublished version ($LATEST).

darrenswhite:
> Hi! I wanted to confirm, does this include publishing versions? I have heard from AWS that versions are required for full use of Snap Start.
> 
> See: https://docs.aws.amazon.com/lambda/latest/dg/snapstart.html
> 
> > You can use SnapStart only on [published function versions](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html#configuration-versions-config) and [aliases](https://docs.aws.amazon.com/lambda/latest/dg/configuration-aliases.html) that point to versions. You can't use SnapStart on a function's unpublished version ($LATEST).

this PR does not include publishing versions. you can use the already existing variable `publish` to enable publishing a new version here https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/master/variables.tf#L125

darrenswhite:
Replaced with new PR #406.

darrenswhite:
@mmichna I have updated the new PR with your comments.

@faxx1080 the new PR will automatically publish a new version when snap start is enabled.

faxx1080:
Hi @darrenswhite 

Thank you! However I am not sure what this does, we can already use "publish = true" for snapstart.

My thought is what happens to older versions of the function?

Currently I wrote a Perl script that uses the AWS CLI to enumerate versions not tagged and deletes those one by one. I use a null terraform provider to invoke the Perl script.

PR #395, opened 2023-01-10
feat: Expose platform argument to the docker-build module

mbeacom:
## Description

The purpose of this PR is to add support for multi-arch or targeted arch image builds.

## Motivation and Context

Given the prevalence of multi-arch image builds these days, it's likely a good idea to expose platform target via the module.  This allows a user to define, for example: `linux/amd64` or `linux/amd64,linux/arm64`

## Breaking Changes

None

## How Has This Been Tested?

### DIFF
--- a/modules/docker-build/variables.tf
+++ b/modules/docker-build/variables.tf
@@ -76,3 +76,9 @@ variable "keep_remotely" {
   type        = bool
   default     = false
 }
+
+variable "platform" {
+  description = "The target architecture platform to build the image for."
+  type        = string
+  default     = null
+}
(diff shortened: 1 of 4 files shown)


modules/docker-build/variables.tf
```diff
@@ -76,3 +76,9 @@ variable "keep_remotely" {
   type        = bool
   default     = false
 }
+
+variable "platform" {
+  description = "The target architecture platform to build the image for."
+  type        = string
```
bryantbiggs, line 82: a bit odd that it wants a string instead of an array of strings, but alas https://github.com/kreuzwerker/terraform-provider-docker/blob/617f2ed44e45d45b07161d3f194c3c324ef0511e/internal/provider/resource_docker_registry_image.go#L315

PR #404, opened 2023-01-24
chore: Upgrade CI workflows to use non-deprecated runtimes

dev-slatto:
## Description

The `Release` and `Lock` workflow are using an old version and hence a [deprecated runtime (Node.js 12)](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/). Bumping these workflows will resolve the deprecation warning in the affected workflows.

I've not addressed the warnings on the `Tests` workflow as Github still hasn't released updated docs on how to reference matrix after set-output is gone. My guess is that this warning is invalid for these references to `matrix.x`. If anything changes I'll add a PR on that one as well. 

## Motivation and Context

Make sure that CI workflows uses non-deprecated runtimes.

## Breaking Changes

None.

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -8,7 +8,7 @@ jobs:
   lock:
     runs-on: ubuntu-latest
     steps:
-      - uses: dessant/lock-threads@v3
+      - uses: dessant/lock-threads@v4
         with:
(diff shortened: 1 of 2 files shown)


antonbabenko:
This PR is included in version 4.9.0 :tada:

Issue #418, opened 2023-02-10
Support for replace_security_groups_on_destroy and replacement_security_group_ids


ricoli:
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅: [v4.54.0](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v4.54.0)

## Is your request related to a problem? Please describe.
Yes - destroying VPC associated lambdas taking a long time because of waiting for security group to no longer be used by ENIs. See https://github.com/hashicorp/terraform-provider-aws/issues/10329#issuecomment-1425914496 for details

## Describe the solution you'd like.
Support for the newly introduced `replace_security_groups_on_destroy` and `replacement_security_group_ids` lambda function arguments

antonbabenko:
Good idea! PR is welcome.

PR #422, opened 2023-02-13
feat: Added configuration options to replace security groups on destroy of Lambda function

ricoli:
## Description
Add support for the newly introduced `replace_security_groups_on_destroy` and `replacement_security_group_ids` lambda function arguments in [v4.54.0](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v4.54.0)

## Motivation and Context
Implements https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/418

## Breaking Changes
None

## How Has This Been Tested?


variables.tf
```diff
@@ -223,6 +223,18 @@ variable "snap_start" {
   default     = false
 }
 
+variable "replace_security_groups_on_destroy" {
+  description = "(Optional) When true, all security groups defined in vpc_security_group_ids will be replaced with the default security group after the function is destroyed. Set the replacement_security_group_ids variable to use a custom list of security groups for replacement instead."
+  type        = bool
+  default     = false
```
antonbabenko, line 229:
```suggestion
  default     = null
```

Let Terraform AWS provider decide which default value is best.

PS: I don't think there is an issue (e.g., BC break) if we use `true` as default. It doesn't seem to me that it will have bad consequences for users.

antonbabenko (changes requested):
LGTM. Minor improvement and docs update.

variables.tf
ricoli: good shout, indeed it's best to leave the decision of a default to the provider :+1:

ricoli: I've pushed the change

antonbabenko:
This PR is included in version 4.12.0 :tada:

nils-at-slashwhy:
Hi Ricardo and Anton. Are you sure that you introduced this feature like intended?

The docs of the lambda module say that the attributes replace_security_groups_on_destroy and replacement_security_group_ids are optional.

But when I use the new version of the module, without changing my terraform code, my terraform deploy fails:

> Error: Missing required argument
> ...
> │ "replacement_security_group_ids": all of
> │ `replace_security_groups_on_destroy,replacement_security_group_ids` must be
> │ specified

antonbabenko:
@nils-at-slashwhy #433 is already about it. I will look at it during the next couple of hours and fix it.

ricoli:
> Hi Ricardo and Anton. Are you sure that you introduced this feature like intended?
> 
> The docs of the lambda module say that the attributes replace_security_groups_on_destroy and replacement_security_group_ids are optional.
> 
> But when I use the new version of the module, without changing my terraform code, my terraform deploy fails:
> 
> > Error: Missing required argument
> > ...
> > │ "replacement_security_group_ids": all of
> > │ `replace_security_groups_on_destroy,replacement_security_group_ids` must be
> > │ specified

Oh no, I wonder if replacement_security_group_ids should also be `null` as default then as to not trigger this validation?

antonbabenko:
@ricoli Probably. Please try it if you have time.

ricoli:
> @ricoli Probably. Please try it if you have time.

I'm on it - sorry about the trouble!

ricoli:
> > @ricoli Probably. Please try it if you have time.
> 
> I'm on it - sorry about the trouble!

Indeed switching the default of `replacement_security_group_ids` to null prevents the argument validation from firing, PR is at https://github.com/terraform-aws-modules/terraform-aws-lambda/pull/434

PR #429, opened 2023-03-06
fix: Add if condition for empty hash_extra

chtakahashi:
### EDIT
Thanks to @alabonne-ca for pointing out that this issue is resolved upstream at the provider level here: https://github.com/hashicorp/terraform-provider-external/issues/193. It still could be worth merging since it would prevent this issue from resurfacing down the line if this were to come up again?

### Original Post
We suddenly ran into an issue in our work where we get the error:
```
Error: External Program Execution Failed
```
where the root cause of the error was:
```
│     content_hash.update(hash_extra.encode())
│                         ^^^^^^^^^^^^^^^^^
│ AttributeError: 'NoneType' object has no attribute 'encode'
```
We do not use the `hash_extra` variable, and looking in the code we noted that it was the only variable which is able to be the empty string on which the Python `str.encode()` function is used. In order to circumvent this issue, I simply added an if condition to make sure that it was populated before running the `.encode()`.

## Breaking Changes
It should not.

## How Has This Been Tested?
Only local testing

### DIFF
--- a/package.py
+++ b/package.py
@@ -1430,7 +1430,8 @@ def prepare_command(args):
     content_hash = bpm.hash(hash_extra_paths)
     content_hash.update(json.dumps(build_plan, sort_keys=True).encode())
     content_hash.update(runtime.encode())
-    content_hash.update(hash_extra.encode())
+    if hash_extra:
+        content_hash.update(hash_extra.encode())
     content_hash = content_hash.hexdigest()
 
     # Generate a unique filename based on the hash.


plukevdh:
This is breaking us as well. Explicitly setting that param resolves, but something like this is probably the better fix.

alabonne-ca:
Hi all, this appeared to be a bit more upstream, see [here](https://github.com/hashicorp/terraform-provider-external/issues/193). With the latest provider version this issue was fixed for us.

plukevdh:
Same. The upstream issue and subsequent release resolved for us. Thanks to @alabonne-ca for pointing us that way.

chtakahashi:
closing since the root issue has been resolved and nobody has suggested we merge this one :)

PR #453, opened 2023-04-25
feat: Rebuild deployment package on architecture change

Dunedan:
## Description

Rebuild the deployment package on CPU architecture change.

## Motivation and Context

Previously the CPU architecture configured for an AWS Lambda Function wasn't taken into account when generating the file name hash for the deployment package. This meant that changing the CPU architecture of an AWS Lambda Function did not trigger a rebuild of the deployment package. As deployment packages might contain CPU-specific binaries, this could've caused deployments breaking AWS Lambda Functions.

## Breaking Changes

As the file name hash changes due to these changes, a rebuild of all deployment packages will happen once using a version of this Terraform module with the changes included.

Also not really a breaking change, but a minor annoyance of the current implementation is that switching from an unconfigured CPU architecture to `x86-64` will also result in a rebuild of the deployment package, as that results in a switch of the value for the architecture from `null` to `["x86-64"]`. However, I'm not sure what'd be the best way to fix that and as it only happens when intentionally changing the architecture attribute, I believe it's fine as is as well.

### DIFF
--- a/package.tf
+++ b/package.tf
@@ -28,7 +28,7 @@ data "external" "archive_prepare" {
 
     artifacts_dir            = var.artifacts_dir
     runtime                  = var.runtime
-    compatible_runtimes      = var.compatible_runtimes
+    compatible_runtimes      = jsonencode(var.compatible_runtimes)
     architectures            = jsonencode(var.architectures)
     compatible_architectures = jsonencode(var.compatible_architectures)
     source_path              = jsonencode(var.source_path)
--- a/package.tf
+++ b/package.tf
@@ -28,6 +28,7 @@ data "external" "archive_prepare" {
 
     artifacts_dir = var.artifacts_dir
     runtime       = var.runtime
+    architectures = jsonencode(var.architectures)
     source_path   = jsonencode(var.source_path)
     hash_extra    = var.hash_extra
     hash_extra_paths = jsonencode(
--- a/package.tf
+++ b/package.tf
@@ -26,11 +26,13 @@ data "external" "archive_prepare" {
       docker_entrypoint         = var.docker_entrypoint
     }) : null
 
-    artifacts_dir = var.artifacts_dir
-    runtime       = var.runtime
-    architectures = jsonencode(var.architectures)
-    source_path   = jsonencode(var.source_path)
-    hash_extra    = var.hash_extra
+    artifacts_dir            = var.artifacts_dir
+    runtime                  = var.runtime
+    compatible_runtimes      = var.compatible_runtimes
+    architectures            = jsonencode(var.architectures)
+    compatible_architectures = jsonencode(var.compatible_architectures)
+    source_path              = jsonencode(var.source_path)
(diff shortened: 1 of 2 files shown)


Dunedan:
not stale

antonbabenko (changes requested):
Thank you for this addition, but let's make it a bit better by taking into account cases where Lambda Layer is being built.

This means that the user specifies `var.compatible_architectures` instead of `var.architectures`.

Also, let's do a similar change for `var.runtime` -> `var.compatible_runtimes`.

Currently, the build plan phase is broken for Lambda Layers (it doesn't use the correct runtime value in plan jsonfile):

https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/dc9356a4804b98894c5d1bf651e88f5eacdb90d8/examples/complete/main.tf#L205-L216

Dunedan:
`var.compatible_architectures` and `var.compatible_runtimes` get considered for rebuilding the deployment package now as well.

package.tf
```diff
@@ -26,10 +26,13 @@ data "external" "archive_prepare" {
...
     }) : null
 
-    artifacts_dir = var.artifacts_dir
-    runtime       = var.runtime
-    source_path   = jsonencode(var.source_path)
-    hash_extra    = var.hash_extra
+    artifacts_dir            = var.artifacts_dir
+    runtime                  = var.runtime
+    compatible_runtimes      = var.compatible_runtimes
+    architectures            = jsonencode(var.architectures)
```
antonbabenko, line 32:
```suggestion
    architectures            = var.architectures
```

Right? I didn't run the code, but I think it should not be `jsonencoded` everywhere except `sourch_path`, whic has a very dynamic structure (by design).

Should this be applied to `compatible_architectures`, too?

antonbabenko: I didn't run this PR, so I don't know if the proposed change will work. Please try it yourself to make sure that PR works as you expect :)

Dunedan: Sorry, for the delayed response, but I wanted to thoroughly look at this PR again before getting back to you and now finally found the time to do so.

Without using `jsonencode()` it doesn't work, as the `external` data source expects attributes of the `query` attribute to be strings (https://registry.terraform.io/providers/hashicorp/external/latest/docs/data-sources/external), while the values provided here are lists of strings. I've fixed that now, by adding the missing `jsonencode()` for `var.compatible_runtimes`.

While testing the fix, I also noticed that deployment packages would still be rebuilt, if the order of items in the list of values for the attributes changed. I tried to avoid that originally as well, but failed to recognize that the `sort_keys` of `json.dumps()` only affects dictionaries and not lists as well. I fixed that as well, so changing the order of items in the list of values for the attributes doesn't cause a rebuild of the deployment package either anymore.

From my perspective this PR is good to go again now.

Dunedan:
Not stale, just waiting for @antonbabenko to have a look at it again.

Dunedan:
still relevant

PR #462, opened 2023-05-16
feat: Added control to create logs by Lambda@Edge in all regions

rysi3k:
## Description
Added simple flag, for not create IAM policy which allows to create logs by Lambda@Edge in regions other than N.Virginia.

## Motivation and Context
In some cases that logs are not necessary, but storing it costs extra $.

## Breaking Changes
NO

## How Has This Been Tested?


examples/complete/main.tf
```diff
@@ -262,6 +262,7 @@ module "lambda_at_edge" {
   source = "../../"
 
   lambda_at_edge = true
+  create_lambda_at_edge_all_regions_access_policy = false
```
antonbabenko, line 265:
```suggestion
```

antonbabenko (changes requested):
I like the idea of restricting logs by region, but I propose changing the name of the variable to something like `lambda_at_edge_logs_all_regions`.

Also, please remove the changes from `examples/complete/main.tf` as it is a very minor case.

Issue #455, opened 2023-05-05
Add support for build, package & deploy of Golang + arm64 + provided.al2


vainkop:
## Is your request related to a new offering from AWS?
No

## Is your request related to a problem? Please describe.

It doesn't look possible to use the current implementation of the module & `package.py` to build, package & deploy a `Golang` lambda for `arm64` for `provided.al2` runtime (which is the only run time at the moment working with arm64 Golang)

## Describe the solution you'd like.
To be able to configure build, package & deploy for Golang + arm64 + provided.al2 runtime out of the box.

It is probably a good idea to make it possible to fully redefine the commands used for the build inline + be able to use external Makefile regardless of the runtime & architecture.

A simple example of a Golang + arm64 build + package:
```
env GOOS=linux GOARCH=arm64 go build -o bootstrap src/main.go
zip -j deploy.zip bootstrap
```

https://docs.aws.amazon.com/lambda/latest/dg/golang-package.html

https://aws.amazon.com/blogs/compute/migrating-aws-lambda-functions-to-arm-based-aws-graviton2-processors/

https://github.com/mthenw/serverless-go-plugin/blob/master/README.md#how-to-run-golang-lambda-on-arm

vi2co:
+1 to keep this issue open

villekr:
ARM64 AL2 would be also interesting. How would one use this module for deploying Go Lambda for X86_64 architecture? Simple example would be nice. I've used this module so far with Python and Node.js with layers so will start to look also using this with Go.

vi2co:
@villekr 
I would like to use it for [Loki Promtail Lambda](https://github.com/grafana/loki/tree/main/tools/lambda-promtail/lambda-promtail) (X86_64 or ARM64).
So far I just put compiled Lambda archive into my Terraform repo. But I would like to automate this process with Terraform e.g. download the source code from Github, build it and deploy Lambda.

vainkop:
It looks pretty clear that to add Golang builds support one needs to either work on [package.py](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/v4.18.0/package.py#L776) or add a condition into [package.tf](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/v4.18.0/package.tf#L8) which will trigger a different script for Golang + provided.al2 runtime.
Too bad I don't have time to implement it at the moment, hopefully I will find it soon.

PR #490, opened 2023-08-02
feat: Build and deploy go for arm64

lexhuismans:
## Description

This PR builds the go code before zipping and deploying when the runtime is `provided.al2` and the code file ends with `.go` and the architecture is `arm64`. Might be a start for solving #455.

### Desired improvements
Some improvements I would like to add with additional input.

- Currently it is hardcoded for `arm64` architecture  I want this to be set programmatically.
- No Docker support.
- Go files are also included in the Zip (not only the binary).

## Motivation and Context

With this change the process of building the go code for `provided.al2` runtime is automated. 

Partly solves #455. 

## Breaking Changes

I don't think it breaks anything. Only thing that is added are some additional cases in the if-statements.

## How Has This Been Tested?

### DIFF
--- a/package.py
+++ b/package.py
@@ -714,6 +714,19 @@ class BuildPlanManager:
 
                 step('npm', runtime, requirements, prefix, tmp_dir)
                 hash(requirements)
+        
+        def go_build_step(path, prefix=None, tmp_dir=None):
+            command = "go"
+            if not os.path.isfile(path):
+                raise RuntimeError('File not found: {}'.format(path))
+            else:
+                if not query.docker and not shutil.which(command):
+                    raise RuntimeError(
+                        "Go command not found should be "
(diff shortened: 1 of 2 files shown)


package.py
```diff
@@ -882,6 +897,16 @@ def execute(self, build_plan, zip_stream, query):
                         else:
                             # XXX: timestamp=0 - what actually do with it?
                             zs.write_dirs(rd, prefix=prefix, timestamp=0)
+            elif cmd == "go":
+                path, tmp_dir = action[1:]
+                with build_go_binary(query, path, tmp_dir) as rd:
+                    if rd:
+                        if pf:
```
lexhuismans, line 904: Looks like pf is always `None`(?) so we might remove all these if statements?

PR #494, opened 2023-08-09
feat!: Disable creation of SAM metadata null-resources by default

antonbabenko:
There is simply not much value for most users in having SAM metadata resources created, so let's disable them by default.


antonbabenko:
This PR is included in version 6.0.0 :tada:

Issue #414, opened 2023-02-03
local-exec step fails when trying to run 'npm install'  on Windows


james-mwakichako:
## Description

Trying to deploy a lambda on Windows and getting the following error:

`
module.lambdaToDynamo_lambda.null_resource.archive[0] (local-exec):     raise RuntimeError(
module.lambdaToDynamo_lambda.null_resource.archive[0] (local-exec): RuntimeError: Nodejs interpreter version equal to defined lambda runtime (nodejs16.x) should be available in system PATH
`
I have verified that I have node and npm installed . 
<img width="609" alt="Screenshot 2023-02-03 at 3 19 08 PM" src="https://user-images.githubusercontent.com/93148349/216712659-9958d593-ac88-4d45-84c1-06f04218f3c0.png">

<img width="590" alt="Screenshot 2023-02-03 at 3 22 09 PM" src="https://user-images.githubusercontent.com/93148349/216713153-c521c1f2-2a5b-4927-9801-4a285f2d74fa.png">

I have ran into the same error on GitBash and CMD.  NOT seeing this issue on Mac. 

Below are the system specs:
Terraform : v1.3.7
Edition  Windows 10 Enterprise
Version 20H2
Installed on         ‎09/‎01/‎2022
OS build               19042.2486
Experience         Windows Feature Experience Pack 120.2212.4190.0

## Versions

- Module version [Required]: 4.7.1

- Terraform version: v1.3.7
- Provider version(s): v4.47.0
- Python version  3.10

## Reproduction Code 

```
module "lambdaToDynamo_lambda" {
  source  = "terraform-aws-modules/lambda/aws"
  version = "4.7.1"
 
  function_name = "testing-lambda"
  handler       = "index.handler"
  runtime       = "nodejs18.x"
 
  source_path = "../lambda/src/lambdaToDynamo"
 
  memory_size = 128
  timeout     = 60
 
  cloudwatch_logs_retention_in_days = 7
 
  tags = local.common_tags
}
 
output "lambdaToDynamo_lambda_function_name" {
  value = module.lambdaToDynamo_lambda.lambda_function_name
}
 
output "lambdaToDynamo_lambda_role_arn" {
  value = module.lambdaToDynamo_lambda.lambda_role_arn
}
 
 
provider "aws" {
  region = var.region
}
 
terraform {
 
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "4.47.0"
    }
  }
}
 
[...]

PR #502, opened 2023-10-05
fix: Fixed npm install on Windows without having to use wsl

eicki:
## Description

Use of npm.cmd on WINDOWS instead of npm within install_npm_requirements method

## Motivation and Context

https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/414 describes the issue, it has been closed with the resolution that WSL should be used. While this is a solution not all development teams use WSL. Additionally, the package.py already contains specific WINDOWS handling anyway (e.g. for python).

Fixes https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/414

## Breaking Changes

no breaking change imho

## How Has This Been Tested?

locally on my Windows machine.

### DIFF
--- a/package.py
+++ b/package.py
@@ -1240,12 +1240,16 @@ def install_npm_requirements(query, requirements_file, tmp_dir):
         shutil.copyfile(requirements_file, target_file)
 
         subproc_env = None
-        if not docker and OSX:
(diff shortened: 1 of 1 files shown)


antonbabenko:
This PR is included in version 6.0.1 :tada:

PR #508, opened 2023-10-24
feat: Allows tags to be provided only to the function

mike-carey:
## Description

Adds a `function_tags` input to allow function-only tags to be passed.

## Motivation and Context

We would like to configure tags that are specific to the lambda function and not any other resources created under this module.

## Breaking Changes

No

## How Has This Been Tested?
- [X] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)

We have also tested using this feature against our internal infrastructure.


antonbabenko:
This PR is included in version 6.1.0 :tada:

PR #509, opened 2023-10-31
feat: Add source_code_hash var

Octogonapus:
## Description

This PR adds a `source_code_hash` variable which overrides the computed hash using a user-supplied value.

## Motivation and Context

This change is needed when you are using a lambda zip which is created by Terraform.
Terraform's plan will pick up the hash from an existing zip file.
If the zip file is recreated at apply time, the hash will change and the plan will become inconsistent, leading to an error like this:
```
╷
│ Error: Provider produced inconsistent final plan
│ 
│ When expanding the plan for module.foo_lambda.module.lambda.aws_lambda_function.this[0] to include new values learned so far during
│ apply, provider "registry.terraform.io/hashicorp/aws" produced an invalid new value for .source_code_hash: was cty.StringVal("2v4cb4VO4El5iGo7/SQ3bLn/lEcSUJAI8o4FBB8u6w0="), but now
│ cty.StringVal("MXbLA/B/BhPwZpVrpD81N27nytZJJ9lXxylwWW5aL0M=").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

There are lots of issues in the AWS provider like this but I'm not sure if they are caused by this exact use case so I won't link them here.

## Breaking Changes

No breaking changes; the source code hash is still computed by default.

## How Has This Been Tested?

I tested this in my own project.


Octogonapus:
you are right this can be solved another way

Issue #503, opened 2023-10-11
Add variable to manage the CreateLogGroup permission


jake-naughton:
## Is your request related to a new offering from AWS?
- Already exists.

## Is your request related to a problem? Please describe.
I have to manage IAM roles outside of this module because it enforces `logs:CreateLogGroup` permission when the module provisions the CloudWatch log group. However, there are many cases where you don't need or want this permission attached to your lambda and it is a nuisance when you are dealing with ephemeral environments. The problem is described in this issue:
https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/342

## Describe the solution you'd like.
To add a variable that allows for the `logs:CreateLogGroup` to be turned off even when the module is responsible for provisioning the CloudWatch logs. By default it would be on so no change in behaviour.

```
variable "attach_create_log_group_permission" {
  description = "..."
  type             = bool
  default        = true
}

```

```
data "aws_iam_policy_document" "logs" {
  count = local.create_role && var.attach_cloudwatch_logs_policy ? 1 : 0

  statement {
    effect = "Allow"

    actions = compact([
      !var.use_existing_cloudwatch_log_group && var.attach_create_log_group_permission ? "logs:CreateLogGroup" : "",
      "logs:CreateLogStream",
      "logs:PutLogEvents"
    ])

    resources = flatten([for _, v in ["%v:*", "%v:*:*"] : format(v, local.log_group_arn)])
  }
}
```

## Describe alternatives you've considered.
Creating my own IAM role outside of the module and passing it in. Could also create the CloudWatch log group outside of the module.

## Additional context
Happy to put up a PR for this improvement if it is acceptable.

KamilBlaz:
Hello, can I take care of it?

jake-naughton:
Any thoughts on this?

antonbabenko:
Hi @jake-naughton !

Please make a PR with the proposed change.

jake-naughton:
Hi @antonbabenko, I have raised the PR here: https://github.com/terraform-aws-modules/terraform-aws-lambda/pull/514

PR #514, opened 2023-11-15
feat: Added variable to control the create log group permission

jake-naughton:
## Description
Adding a variable to control the create log group permission on the lambda role. Default for the new variable is to add the permission to the role.

## Motivation and Context
Described in the following feature request:
Fixes https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/503

## Breaking Changes
None, variable default will keep current behaviour the same.

## How Has This Been Tested?

Tested running the example/complete module with init, plan, apply, destroy and inspected the iam roles to confirm that existing lambda roles still had the create log group permission and the new lambda role does not have it.


iam.tf
```diff
@@ -122,7 +122,7 @@ data "aws_iam_policy_document" "logs" {
     effect = "Allow"
 
     actions = compact([
-      !var.use_existing_cloudwatch_log_group ? "logs:CreateLogGroup" : "",
+      !var.use_existing_cloudwatch_log_group && var.attach_create_log_group_permission ? "logs:CreateLogGroup" : "",
```
antonbabenko, line 125: If you want to reuse the existing cloudwatch log group (from previous runs), set `use_existing_cloudwatch_log_group = true`.

I am not sure that we need to have an identical variable `attach_create_log_group_permission`.

Right?

jake-naughton: The scenario I am dealing with is ephemeral environments. When I first create a new environment the cloudwatch log group does not exist. So I need `use_existing_cloudwatch_log_group` set to false. When I destroy the environment, the log group may or may not persist. I then want to recreate that environment at a later date which, if the group exists, will fail. 

`use_existing_cloudwatch_log_group` only works if you know the CloudWatch group will be there, I do not.

As stated in the feature request, a work around would be to create the CloudWatch group separately. However, I would rather contain the logic to a single module.

jake-naughton: To be honest, since we provision the CloudWatch group separately, having `CreateLogGroup` on the Lambda seems redundant to me unless there is something I'm missing.

jake-naughton: @antonbabenko thoughts?

antonbabenko (approved):
Looks good to be merged. Thanks!

antonbabenko:
This PR is included in version 6.5.0 :tada:

PR #526, opened 2024-01-04
chore: Update to supported versions of Python and bump pytest

n8felton:
## Description
Python 3.7 is EoL as of 2023-06-27 and AWS Lambda has dropped support for Python 3.7.

Supported lambda versions include 3.8 - 3.12
https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html

## Motivation and Context
Support the latest versions of Python that are supported by AWS Lambda
https://devguide.python.org/versions/
https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html


antonbabenko:
Thanks @n8felton !

antonbabenko:
This PR is included in version 6.8.0 :tada:

PR #527, opened 2024-01-06
docs: Update docs to clarify create_role and lambda_role usage

victortrac:
## Description

Adds clarification to the README and variables.tf about create_role and lambda_role. 

## Motivation and Context

The lack of this information in the docs made troubleshooting this module take some time.

## Breaking Changes

None.

## How Has This Been Tested?


antonbabenko:
Such an addition to the explanation is not necessary because if we change the logic of the module (rarely we do it) we need to remember to update the description of the variable. Instead, we should provide good feature coverage with examples.

PR #521, opened 2023-12-13
feat: Add control to use timestamp to trigger the package creation or not (useful for CI/CD)

samuel-phan:
## Description

Add a variable `trigger_on_package_timestamp` (boolean - defaults to `true`) to whether use timestamp to trigger the package creation or not.

### Problem

1. Use the lambda module to create a lambda. Run `terraform apply`. The lambda is created.
2. Delete the `builds` directory (to simulate a CICD job with an empty environment).
3. Run `terraform plan` again, it will detect a drift because the timestamp has changed.

One solution was to add the variable `recreate_missing_package = false`, but when the CICD job (without any `builds` directory) runs `terraform apply` will face this error:

```
│ Error: Provider produced inconsistent final plan
│
│ When expanding the plan for module.foo_lambda.aws_lambda_function.this[0] to include new values learned so far during apply, provider "registry.terraform.io/hashicorp/aws"
│ produced an invalid new value for .source_code_hash: was cty.StringVal("cqn6fMnsZk+KB1YcvY+mV9lVQvKZ7T7n+vpBjv/qg3s="), but now
│ cty.StringVal("mM6yOCDC4GsOEw8Sqyb7t1aZPQqbwu/g27igFYoF6ko=").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

### Solution

Add a variable `trigger_on_package_timestamp` (boolean - defaults to `true`) to whether use timestamp to trigger the package creation or not.

Fix #396

## Breaking Changes

None.

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

I have added an `example/simple-cicd` with a `test.sh` inside to run the test scenario.

PR #529, opened 2024-01-11
feat: make the archive resource change when its content change

davidvasandani:


antonbabenko:
I don't think that this is so much releavant since #521 was merged.

Issue #492, opened 2023-08-05
Ignore changes to `image_uri`


jensenbox:
Bringing this up again as it is a good feature.

Something like what was done for the EC2 image AMI ignore would be ideal: https://github.com/terraform-aws-modules/terraform-aws-ec2-instance/issues/325#issuecomment-1527640607

Issue #480, opened 2023-07-05
Add suppurt f


ulielitay:
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅: please list the AWS provider version which introduced this functionality: >= 3.75.2
  -

## Is your request related to a problem? Please describe.
I'm using ecr images, each time i want to run the TF the ecr image + environment variables are changed to the TF one and not the already exist by the pipeline

## Describe the solution you'd like.
add life cycle to the resources so we can "one time deploy" (or add the option to add to the module from our side what to ignore
https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle

i have it in my own personal module i've created

  environment {
    variables = var.environment_variables
  }

  lifecycle {
    ignore_changes = [
      environment,
      filename,
      source_code_hash,
      source_code_size,
      last_modified,
      image_uri
    ]
  }

antonbabenko:
Terraform doesn't support the usage of variables inside of `lifecycle` `ignore_changes`, so we can't make it part of the module.

Issue #213, opened 2021-10-08
Allow to ignore external changes to image_uri


marianzange:
## Is your request related to a problem? Please describe.

It's currently not possible to ignore changes to the `image_uri` field. The use case is quite simple: The Lambda function is initially deployed through Terraform, referencing the `LATEST` tag on the docker repo. Subsequent deploys are done through our CI/CD pipeline which simply changes the `image_uri` of the Lambda via AWS API to the current stable tag.

Here's an example diff I get, after the image has been bumped by our deployment process:
```
# module.*******.aws_lambda_function.this[0] will be updated in-place
~ resource "aws_lambda_function" "this" {
      id                             = "FUNCTION_NAME"
    ~ image_uri                      = "********.dkr.ecr.eu-west-1.amazonaws.com/IMAGE_NAME:8a4c964c9" -> "********.dkr.ecr.eu-west-1.amazonaws.com/IMAGE_NAME:latest"
    ~ last_modified                  = "2021-10-08T09:38:22.000+0000" -> (known after apply)
      tags                           = {}
      # (16 unchanged attributes hidden)
      # (2 unchanged blocks hidden)
  }
```

## Describe the solution you'd like.

Add a parameter to ignore changes in the `image_uri`, e.g. `ignore_image_uri_change = true`

Issue #212, opened 2021-10-07
Support lifecycle ignore changes for image_uri


giom-l:
All my lambdas are deployed as "dummy" lambda so that Terraform creates them and my CI handle the update of images running.
To be able to do that and not trigger a modification everytime terraform is run, I need terraform to set `ignore_changes` lifecycle set on `image_uri`.

I know that lifecycle support has already been required for modules (see https://github.com/hashicorp/terraform/issues/27360).
We couldn't also not make this module capable of handling all kind of lifecycle due to the number of resources handled.

However, there is an option to ignore updates on source_code_hash `ignore_source_code_hash` that let users get this ignore behaviour for other sources than container images.

Would it be possible to use this variable to also set 
```
lifecycle {
    ignore_changes = [
      image_uri
    ]
  }
```
when its value is `true` and get the same behaviour whatever kind of source we use ? 

That would be a minor change and it would'nt be (unless I'm wrong) with lifecycle support on module when it will be released.

zamialloy:
Since [hashicorp/terraform#27360](https://github.com/hashicorp/terraform/issues/27360) is still not out, one solution that could work for everyone is adding a bool variable `ignore_image_uri` and using that to create a new `aws_lambda_function` resource. This resource will be identical to the current `aws_lambda_function.this` but the differences would be in the count and lifecycle policy.
```
resource "aws_lambda_function" "image_function" {
  count = local.create && var.create_function && !var.create_layer && var.ignore_image_uri ? 1 : 0
  lifecycle {
    ignore_changes = [image_uri]
  }
  ... (everything else exactly the same as aws_lambda_function.this)
  ...
}
```
This will ensure the new resource would only be created if that variable is specified.
All references to `aws_lambda_function.this[0]` would need to be updated. A local variable could be added as follows:
```
locals {
  ...
  lambda_used = var.ignore_image_uri ? aws_lambda_function.this[0] : aws_lambda_function.image_function[0]
} 
```
And all references to `aws_lambda_function.this[0]` can be replaced with `local.lambda_used` in main.tf, outputs.tf.

I have tried this out and it works.

The main drawback is that updates to the lambda resource would need to happen in two places going forward. Inside `aws_lambda_function.this` and `aws_lambda_function.image_function`
The other drawback is that if people were previously not using this variable and set it to true, they would need to use `terraform state mv` to migrate the resource to the correct block.

Let me know if this makes sense and is something we can move forward with.

PR #532, opened 2024-01-16
fix: Workaround to allow ignoring changes to `image_uri`

lebenitza:
## Description
Duplicated the whole lambda resource in order to have one with:
```hcl
  lifecycle {
    ignore_changes = [
      image_uri
    ]
  }
```
This is an workaround to https://github.com/hashicorp/terraform/issues/27360 and would fix:
* https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/492
* https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/480
* https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/213
* https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/212

## Motivation and Context
Doing continuous delivery from terraform is a bit tedious. We use lambda with container images but we update those image outside of the terraform state using other toolings and automations.

This kind of workaround already exists in other aws terraform modules.

## Breaking Changes
No breaking change. When `ignore_image_uri` is set to `true`, it is going to recreate the existing terraform lambda resources!

## How Has This Been Tested?

- [x] We already migrated environments to these changes with no issue, just by setting `ignore_image_uri = true` and applying.

You might not want to accept this at all or in its current form but it might help somebody.


antonbabenko (commented):
I understand the desire to get this problem fixed somehow, but I don't want to increase the complexity of the module two times to support it.

The root problem should be fixed in the Terraform in the first place.

Unfortunately, I can't accept this solution as is. :(

lebenitza:
Totally understandable. The root issue is 3 years old now, I have no faith is going to be fixed any time soon so I'll keep the fork with this ugly workaround in the meantime.

antonbabenko:
Yes, we have to have various workarounds for this. Unfortunately.

PR #538, opened 2024-01-26
feat: Added support to override default tags of provider in S3 object

antonbabenko:
## Description
Inspired by https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/pull/261 to workaround the limitation of 10 tags for S3 objects.


antonbabenko:
This PR is included in version 7.2.0 :tada:

PR #551, opened 2024-03-22
fix: Fixed constant drift with Lambda logging configuration

antonbabenko:
A simple configuration like this was marking `version`, `qualified_arn` and `qualified_invoke_arn` as `(known after apply)`:
```
  logging_log_format = "JSON"
```

The user used to pass some extra arguments to reduce noise:
```
  logging_log_group             = "/aws/lambda/adapting-tadpole-test"
  logging_application_log_level = "DEBUG"
  logging_system_log_level      = "INFO"
```

This PR sets defaults for log level to avoid such behavior.

Ideally, this should be fixed in Terraform AWS provider. Related issues:
https://github.com/hashicorp/terraform-provider-aws/issues/33383
https://github.com/hashicorp/terraform-provider-aws/issues/33955


antonbabenko:
This PR is included in version 7.2.3 :tada:

PR #552, opened 2024-03-25
feat: Allow ignoring changes in image_uri

atorrescogollo:
## Description
Allow to a rollout the image_uri (from a different place, e.g: Jenkins, GitHub Actions, etc) without requiring you to update the terraform code.

Unfortunately, I don't think there is a way to parametrize the ignore_changes field. Though, I wrote a feature request to opentofu to try to address this:
* https://github.com/opentofu/opentofu/issues/1432

## Motivation and Context
I want to create the lambda so that it targets to `latest` tag by default and I have always something that runs. However, when I make a change that only concerns to the code, I don't need to run terraform (which has all the dependencies for the lambda). I just want to change the image_uri to target the newly created image (even with aws-cli).

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -709,7 +709,7 @@ No modules.
 | [aws_iam_role_policy_attachment.vpc](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role_policy_attachment) | resource |
 | [aws_lambda_event_source_mapping.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_event_source_mapping) | resource |
 | [aws_lambda_function.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function) | resource |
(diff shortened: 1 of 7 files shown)


antonbabenko:
I understand this PR's reasoning but don't like the code duplication because it significantly increases the maintenance cost.

Unfortunately, I don't see a way to merge this as-is. I suggest that you deploy as you do and change the value via Terraform configuration to avoid the revert.

atorrescogollo:
Ok, I understand. I uploaded this PR because I saw the same issue in the secrets-manager module: https://github.com/terraform-aws-modules/terraform-aws-secrets-manager/blob/v1.1.2/main.tf#L92-L116 . But I get that here is way more maintenance. I think I'll have to maintain it in my fork.

Then, we'll see if eventually this can be an opentofu feature instead. I post the issue link in case someone wants to support this feature by putting a 👍 in the opentofu issue :
* https://github.com/opentofu/opentofu/issues/1432

PR #561, opened 2024-04-23
feat: Added create before destroy on aws_lambda_permission

jinnko:
## Description

Update lambda permissions handling so there's no interruption to the permissions available to a lambda.

The change keeps the module interface the same and so there's no changes needed to the documentation.

The only material change visible to users is that the statement_id will now be used as a prefix rather than as a specific value.

## Motivation and Context

When any parameters contain dynamic elements, e.g. `data.aws_region.name`, and that is evaluated in an intermediate module, it is not known until apply time, which causes the permission to he removed and added.  During this window, Cloudwatch Logs will see errors when delivering events to the subscription and stop sending events for 10 minutes.

By switching to a statement_id_prefix and using a create_before_destroy lifecycle we avoid such interruptions.

## Breaking Changes

No breaking changes.

## How Has This Been Tested?

  - No updates needed
  - tested with `examples/alias` which is confirmed to use both `module.lambda_function.aws_lambda_permission.unqualified_alias_triggers["APIGatewayAny"]` and `module.lambda_function.aws_lambda_permission.current_version_triggers["APIGatewayAny"]`

This has been tested via our own module that in turn calls this module.  Prior to this change we saw the terraform traces showing a remove operation followed by an add operation on the lambda permissions.  After this change a new lambda permission is created first, followed by removal of the old permission.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -273,13 +273,17 @@ resource "aws_lambda_permission" "current_version_triggers" {
   function_name = aws_lambda_function.this[0].function_name
   qualifier     = aws_lambda_function.this[0].version
 
-  statement_id       = try(each.value.statement_id, each.key)
-  action             = try(each.value.action, "lambda:InvokeFunction")
-  principal          = try(each.value.principal, format("%s.amazonaws.com", try(each.value.service, "")))
-  principal_org_id   = try(each.value.principal_org_id, null)
-  source_arn         = try(each.value.source_arn, null)
-  source_account     = try(each.value.source_account, null)
-  event_source_token = try(each.value.event_source_token, null)
+  statement_id_prefix = try(each.value.statement_id, each.key)
+  action              = try(each.value.action, "lambda:InvokeFunction")
+  principal           = try(each.value.principal, format("%s.amazonaws.com", try(each.value.service, "")))
+  principal_org_id    = try(each.value.principal_org_id, null)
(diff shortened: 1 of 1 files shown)


bcdady:
I think I need this fix.

antonbabenko (approved):

antonbabenko:
This PR is included in version 7.3.0 :tada:

antonbabenko:
@bcdady Please confirm that it works as expected. If not, please open a new issue.

plukevdh:
This change is believed to be causing a cyclical dependency issues when attaching additional policies for allowed triggers:

```│ Error: Cycle: module.my_app.module.lambda_function.aws_lambda_permission.unqualified_alias_triggers["events"] (destroy deposed <hash>), module.my_app.module.lambda_function.aws_lambda_permission.current_version_triggers["events"], module.my_app.module.lambda_function.aws_lambda_permission.current_version_triggers["events"] (destroy deposed <hash>), module.my_app.module.lambda_function.aws_iam_role_policy_attachment.additional_many[1] (destroy), module.my_app.module.lambda_function.aws_lambda_function.this[0], module.my_app.module.lambda_function.aws_lambda_permission.unqualified_alias_triggers["events"]```

antonbabenko:
@plukevdh Could you please open a new issue and provide the code snippet to reproduce the issue?

PR #583, opened 2024-06-07
feat: Renamed python3.8-11 to python3.12 in examples, added tag to resources

antonbabenko:
- **feat: Added terraform-aws-modules tag**
- **feat: Renamed python3.[8-11] => python3.12 in examples**


main.tf
```diff
@@ -135,7 +135,11 @@ resource "aws_lambda_function" "this" {
     }
   }
 
-  tags = merge(var.tags, var.function_tags)
+  tags = merge(
```
antonbabenko, line 138: @bryantbiggs WDYT about tags? (the rest of the renaming was automatic)

bryantbiggs: The function def, there won't be any insight into the s3 object, cloudwatch could technically be derived but ok either way. Typically, the fewer the better, and tag the main thing which other things can be derived off of that

bryantbiggs (approved):

main.tf
antonbabenko: Updated in 16b854a

antonbabenko:
This PR is included in version 7.5.0 :tada:

PR #600, opened 2024-07-28
feat: Added the skip_destroy argument for functions

catrielg:
## Description
Added the "skip_destroy" argument for the "function" resource. 
More info from the AWS Provider Doc:
(Optional) Set to true if you do not wish the function to be deleted at destroy time, and instead just remove the function from the Terraform state.

## Motivation and Context
Particularly useful for Lambda@Edge that cannot be destroyed if they are still associated with a CloudFront Distribution.


antonbabenko:
This PR is included in version 7.8.0 :tada:

PR #606, opened 2024-08-01
fix: Fix package.py commands after :zip not being executed

mleziva:
## Description
When using the commands option in source_path, commands after ":zip" were not being executed. This PR resolves that unexpected behavior

## Motivation and Context
The root cause was that the commands loop read all provided commands, but it didn't add the commands after a ":zip" command to the plan file. This was remediated by adding a check after the commands loop exited to add the remaining commands to the steps. 

## Breaking Changes
None

## How Has This Been Tested?


mleziva:
@antonbabenko , can you please review when you get a chance. Thanks!

tests/fixtures/node-app/package-lock.json
```diff
@@ -0,0 +1,115 @@
+{
```
antonbabenko, line 1: Do we need to have this file in the examples? If not, let's simplify it and drop it.

antonbabenko (changes requested):
LGTM.

@pdecat what do you think?

tests/fixtures/node-app/package-lock.json
mleziva: It is required for the `npm ci` command, but I just updated the command to be `npm install` and removed this file to simplify it

antonbabenko (approved):

antonbabenko:
This PR is included in version 7.8.1 :tada:

PR #609, opened 2024-08-19
Switch to name_prefix and allow quieting log out

areeseburnett:
- Use `name_prefix` for all IAM resources
- Allow setting `quiet` parameter for local-exec to limit log output


package.tf
```diff
@@ -70,6 +70,7 @@ resource "null_resource" "archive" {
       "--timestamp", data.external.archive_prepare[0].result.timestamp
     ]
     command = data.external.archive_prepare[0].result.build_plan_filename
+    quiet   = var.quiet_local_exec
```
antonbabenko, line 73: This option is useful in some cases.  PR is welcome if you have time.

PR #617, opened 2024-09-04
feat: Add support for kafka event source config

sergiolc:
## Description
Add support for `self_managed_kafka_event_source_config` and `amazon_managed_kafka_event_source_config` to be able to set custom consumer group ids.

## Motivation and Context
This config is available in lambda function and should be available in alias as well so we can specify custom consumer group ids.

### DIFF
--- a/modules/alias/main.tf
+++ b/modules/alias/main.tf
@@ -161,7 +161,7 @@ resource "aws_lambda_event_source_mapping" "this" {
       consumer_group_id = self_managed_kafka_event_source_config.value.consumer_group_id
     }
   }
-  
+
   dynamic "amazon_managed_kafka_event_source_config" {
     for_each = try(each.value.amazon_managed_kafka_event_source_config, [])
     content {
--- a/modules/alias/main.tf
+++ b/modules/alias/main.tf
@@ -155,6 +155,20 @@ resource "aws_lambda_event_source_mapping" "this" {
     }
   }
 
+  dynamic "self_managed_kafka_event_source_config" {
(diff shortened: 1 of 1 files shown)


modules/alias/main.tf
```diff
@@ -155,6 +155,20 @@ resource "aws_lambda_event_source_mapping" "this" {
     }
   }
 
+  dynamic "self_managed_kafka_event_source_config" {
+    for_each = try(each.value.self_managed_kafka_event_source_config, [])
+    content {
+      consumer_group_id = self_managed_kafka_event_source_config.value.consumer_group_id
```
antonbabenko, line 161:
```suggestion
      consumer_group_id = try(self_managed_kafka_event_source_config.value.consumer_group_id, null)
```

modules/alias/main.tf
```diff
@@ -155,6 +155,20 @@ resource "aws_lambda_event_source_mapping" "this" {
...
+    }
+  }
+
+  dynamic "amazon_managed_kafka_event_source_config" {
+    for_each = try(each.value.amazon_managed_kafka_event_source_config, [])
+    content {
+      consumer_group_id = amazon_managed_kafka_event_source_config.value.consumer_group_id
```
antonbabenko, line 168:
```suggestion
      consumer_group_id = try(amazon_managed_kafka_event_source_config.value.consumer_group_id, null)
```

antonbabenko (changes requested):

antonbabenko (approved):

antonbabenko:
This PR is included in version 7.12.0 :tada:

PR #624, opened 2024-09-30
feat: Ignore last_modified on aws_lambda_function

edgar-castle:
## Description

Adds the lifecycle ignore rule to the lambda function. Without this, it will prompt to re-deploy the lambda function anytime a plan/apply is performed (even without a change to the underlying src files). The behavior now is that it'll only trigger a re-deploy if the src code or the module call is changed.

## Breaking Changes

N/A nothing is using this yet. Will be the first addition to the new release.

## How Has This Been Tested?
I verified this locally, was able to do multiple plan/applies on a project without it wanting to update the lambda resource.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,9 @@ repos:
     hooks:
       - id: terraform_fmt
       - id: terraform_wrapper_module_for_each
+        args:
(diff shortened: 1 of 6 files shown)


edgar-castle:
anddd, I PR'd this to the wrong repo, apologies!

Issue #545, opened 2024-02-20
Unable to use build_in_docker feature with Rancher Desktop or Colima


mdesoeuv:
## Description

The build of a Poetry Project in docker with the lambda module fails. It seems related to a permissions problem with the temporary folder created and bound to the container during the build process.

## Versions

- Module version [Required]: 7.2.1
- Terraform version: v1.5.2
- Docker: 25.0.3, build 4debf411d1
- MacOS: 13.6.3

## Reproduction Code [Required]

- An Docker engine alternative to Docker Desktop (Rancher Desktop or Colima)
- `terraform init`
- `terraform apply`

<details>
    <summary>Project Structure</summary>

```
.
├── builds
│   └── lambda
├── lambda
│   ├── README.md
│   ├── docker
│   │   └── Dockerfile
│   ├── poetry.lock
│   ├── pyproject.toml
│   ├── src
│   │   └── api
│   │       ├── __init__.py
│   │       └── main.py
│   └── tests
│       └── __init__.py
├── main.tf
```

</details>

<details>
    <summary>Terraform</summary>

```
module "lambda_backend" {
  source = "terraform-aws-modules/lambda/aws"

  function_name              = "api-lambda"
  description                = "FastAPI in AWS Lambda"
  handler                    = "api.main.handler"
  publish                    = true
  create_lambda_function_url = true

  source_path = [
    "${path.module}/lambda/src",
    {
      path           = "${path.module}/lambda/pyproject.toml"
      poetry_install = true
    }
  ]

  artifacts_dir = "${path.root}/builds/lambda/"

  build_in_docker = true
  runtime         = "python3.10"
  docker_image    = "build-python3.10-poetry"
  docker_file     = "${path.module}/lambda/docker/Dockerfile"

  cloudwatch_logs_retention_in_days = 3
}

output "lambda_url" {
  value = module.lambda_backend.lambda_function_url
}
```

</details>

<details>
  <summary>Lambda Source Code</summary>

```python
from fastapi import FastAPI
from mangum import Mangum


app = FastAPI()

handler = Mangum(app, lifespan="off")


@app.get("/")
def root():
    return {"Hello World !"}
```

</details>

<details>
  <summary>pyproject.toml</summary>

```
[tool.poetry]
name = "api"
version = "0.1.0"
description = ""
authors = ["mdesoeuvre <mehdi.desoeuvre@gmail.com>"]
readme = "README.md"
[...]

PR #638, opened 2024-11-11
feat: Allow temp dir for poetry docker builds

gilandose:
## Description
Allows optionally specifying a temp dir for poetry to bring parity with npm and pip

## Motivation and Context
Docker CLI on ubuntu default for github actions, does not allow mounting /tmp https://stackoverflow.com/questions/65267251/docker-bind-mount-directory-in-tmp-not-working
This results in a frustrating situation where terraform passes locally but fails with strange error in CI pipeline.

Fixes: #545 

## Breaking Changes
N/A

## How Has This Been Tested?
Added temp dir to one poetry example and left rest untouched to prove not breaking change

### DIFF
--- a/examples/build-package/main.tf
+++ b/examples/build-package/main.tf
@@ -426,7 +426,7 @@ module "lambda_layer_poetry" {
     {
       path           = "${path.module}/../fixtures/python-app-poetry"
       poetry_install = true
-      poetry_tmp_dir = "${path.root}/builds/"
+      poetry_tmp_dir = "${path.cwd}/../fixtures"
     }
   ]
   hash_extra = "extra-hash-to-prevent-conflicts-with-module.package_dir"
--- a/examples/build-package/main.tf
+++ b/examples/build-package/main.tf
@@ -426,6 +426,7 @@ module "lambda_layer_poetry" {
     {
       path           = "${path.module}/../fixtures/python-app-poetry"
       poetry_install = true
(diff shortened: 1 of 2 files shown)


gilandose:
@antonbabenko could I get a review on this when convenient, as this stops me from building layers in github action pipelines

examples/build-package/main.tf
```diff
@@ -426,6 +426,7 @@ module "lambda_layer_poetry" {
     {
       path           = "${path.module}/../fixtures/python-app-poetry"
       poetry_install = true
+      poetry_tmp_dir = "${path.root}/builds/"
```
ahlinc, line 429: Could you change this line similar to the bellow existing in examples? https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/abd5a15fd73ddd6ed36569b94f5f7667d73d5f4a/examples/build-package/main.tf#L40
I think it's better to keep the `builds/` folder with just the module generated content and doesn't give such advice in examples to use it for other purposes.

ahlinc:
@antonbabenko this change looks good to me and my latest changes that did to rework the build plan approach doesn't interfere with this one.

examples/build-package/main.tf
gilandose: updated

antonbabenko:
This PR is included in version 7.18.0 :tada:

PR #639, opened 2024-11-17
fix: Skip broken symlinks on hash computing

ahlinc:
Just fix it


antonbabenko:
This PR is included in version 7.14.1 :tada:

PR #657, opened 2025-01-25
fix: Make default tag `terraform-aws-modules` optional

### LINKED ISSUE #651: No hardcoded tags like terraform-aws-modules = "lambda"

I'd like to be in control of the tags being set. In this case, this module adds an extra tag that IMHO does not add value.  In the PR where it was added, no explanation s given except being a feature...
```
{ terraform-aws-modules = "lambda" }
```

https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/5da19e5827002cd7155d83178e7e5c906e2d0c19/main.tf#L140-L144

This is inside a lambda resource...

Remove it, or make it optional. 

## Describe alternatives you've considered.
Local "patching"

## Additional context
introduced in #583

sebastianczech:
## Description

PR makes default tag `terraform-aws-modules` with value `lambda` optional.

## Motivation and Context

Resolves #651 

## Breaking Changes

PR introduces new variable `include_default_tag`, but it's not breaking change as this variable has defined default value.

## How Has This Been Tested?


variables.tf
```diff
@@ -188,6 +188,12 @@ variable "tags" {
   default     = {}
 }
 
+variable "default_tf_aws_modules_tag" {
```
antonbabenko, line 191:
```suggestion
variable "include_default_tag" {
```

antonbabenko (changes requested):
Thank you! Let's rename the variable.

sebastianczech:
thank you @antonbabenko for quick review and suggestion - I've updated the variable name

antonbabenko (approved):

antonbabenko:
This PR is included in version 7.20.1 :tada:

Issue #676, opened 2025-04-23
Sensitive values in Docker additional options are exposed in the plan


pawelpesz:
## Description

If Docker dependencies build is used and the additional options contain a sensitive value, that value is shown in plain text in the plan.

- [X] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: 7.20.1

- Terraform version: 1.11.4
- Provider version(s):
+ provider registry.terraform.io/datadog/datadog v3.60.0
+ provider registry.terraform.io/hashicorp/aws v5.95.0
+ provider registry.terraform.io/hashicorp/external v2.3.4
+ provider registry.terraform.io/hashicorp/http v3.4.5
+ provider registry.terraform.io/hashicorp/local v2.5.2
+ provider registry.terraform.io/hashicorp/null v3.2.4
+ provider registry.terraform.io/hashicorp/random v3.7.2
+ provider registry.terraform.io/mastercard/restapi v1.20.0
+ provider registry.terraform.io/mongodb/mongodbatlas v1.21.4

## Reproduction Code [Required]

```
module "lambda_dependencies" {
  source  = "terraform-aws-modules/lambda/aws"
  version = "~> 7.20.1"

  create_layer    = true
  create_package  = true
  build_in_docker = true

  docker_additional_options = [
    "-e", "SENSITIVE_VALUE=${sensitive("value")}"
  ]
  ...
}
```

## Expected behavior

The sensitive value should not be shown in the plan.

## Actual behavior

The sensitive value is exposed in the plan.

### Terminal Output Screenshot(s)

```
# module.lambda_dependencies[0].local_file.archive_plan[0] will be created
  + resource "local_file" "archive_plan" {
      + content              = jsonencode(
            {
              + artifacts_dir = "..."
              + build_plan    = [
                  + [
                      + [
                          + "pip",
                          + "python3.13",
                          + "./../../../../../requirements/requirements.txt",
                          + "python",
                          + null,
                        ],
                    ],
                ]
              + docker        = {
                  + docker_additional_options = [
                      + "-e",
                      + "SENSITIVE_VALUE=value",
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ THIS SHOULD BE REDACTED
                    ]
                  + docker_build_root         = ""
                  + docker_entrypoint         = null
                  + docker_file               = ""
                  + docker_image              = ""
                  + docker_pip_cache          = null
[...]

PR #677, opened 2025-04-23
fix: Do not expose output from build command in Docker

pawelpesz:
## Description
If dependencies build is performed using Docker, and additional Docker options are specified, and any of those options contain a sensitive value, the content of the archive file has also to be marked as sensitive, to avoid exposing these sensitive options in the Terraform plan. Unfortunately the `external` provider [doesn't support marking output values as sensitive](https://registry.terraform.io/providers/hashicorp/external/latest/docs/data-sources/external#result-1), so this workaround seems like an only viable solution.

## Motivation and Context
Fixes https://github.com/terraform-aws-modules/terraform-aws-lambda/issues/676.

## Breaking Changes
None.

## How Has This Been Tested?

### DIFF
--- a/package.tf
+++ b/package.tf
@@ -49,7 +49,7 @@ data "external" "archive_prepare" {
 resource "local_file" "archive_plan" {
   count = var.create && var.create_package ? 1 : 0
 
-  content              = var.build_in_docker && anytrue([for option in var.docker_additional_options : issensitive(option)]) ? sensitive(data.external.archive_prepare[0].result.build_plan) : data.external.archive_prepare[0].result.build_plan
+  content              = var.build_in_docker ? sensitive(data.external.archive_prepare[0].result.build_plan) : data.external.archive_prepare[0].result.build_plan
   filename             = data.external.archive_prepare[0].result.build_plan_filename
   directory_permission = "0755"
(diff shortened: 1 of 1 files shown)


pawelpesz:
Hello @antonbabenko, I'd be extremely grateful if you could find time to review this. It's a one-liner 😄

package.tf
```diff
@@ -49,7 +49,7 @@ data "external" "archive_prepare" {
 resource "local_file" "archive_plan" {
   count = var.create && var.create_package ? 1 : 0
 
-  content              = data.external.archive_prepare[0].result.build_plan
+  content              = var.build_in_docker ? sensitive(data.external.archive_prepare[0].result.build_plan) : data.external.archive_prepare[0].result.build_plan
```
antonbabenko, line 52: The best solution would be to hide values ONLY if there are secrets inside. Function [`issensitive`](https://developer.hashicorp.com/terraform/language/functions/issensitive) became available in Terraform 1.8 but this module expects Terraform 1.0.

antonbabenko (approved):

antonbabenko (approved):

antonbabenko:
This PR is included in version 7.20.3 :tada:

Issue #658, opened 2025-01-27
Built in code signing as part of module to support prepared zips


Fran-Rg:
## Is your request related to a new offering from AWS?
  - Yes ✅: please list the AWS provider version which introduced this functionality
All visible on https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/master/examples/code-signing/main.tf#L55

## Is your request related to a problem? Please describe.
To support code signing with this module it's required to pre-zip and upload the file to s3. This causes an issue when we rely on the packaging happening within the module as the zip is produced within the module and can't be signed.

## Describe the solution you'd like.
Add a `aws_signer_signing_job` resource with expected variables:
* [required] enable code signing
* [required] signing profile_name
* [optional] destination s3 prefix otherwise use the same as `var.s3_prefix`
* [optional] destination s3 bucket otherwise use the same bucket as `var.s3_bucket`

## Describe alternatives you've considered.
I tried to use the output `module.this.s3_object.bucket` to sign it and inject it to the `s3_existing_package` but that creates a looping dependency.

## Additional context
Checkov rates code signing as a high finding: https://docs.prismacloud.io/en/enterprise-edition/policy-reference/aws-policies/aws-general-policies/bc-aws-272

antonbabenko:
You are right, there is an example of how to do it, but we currently don't have a submodule that can handle it. We may create it in the future.

Fran-Rg:
@antonbabenko I've raised a PR to support code signing: https://github.com/terraform-aws-modules/terraform-aws-lambda/pull/660

PR #691, opened 2025-07-16
feat: Code signing support #658

Fran-Rg:
## Is your request related to a new offering from AWS?
Yes ✅: please list the AWS provider version which introduced this functionality
All visible on https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/master/examples/code-signing/main.tf#L55
To support code signing with this module it's required to pre-zip and upload the file to s3. This causes an issue when we rely on the packaging happening within the module as the zip is produced within the module and can't be signed.

Add a aws_signer_signing_job resource with expected variables:

[required] enable code signing
[required] signing profile_name
[optional] destination s3 prefix otherwise use the same as var.s3_prefix
[optional] destination s3 bucket otherwise use the same bucket as var.s3_bucket
## Describe alternatives you've considered.
I tried to use the output module.this.s3_object.bucket to sign it and inject it to the s3_existing_package but that creates a looping dependency.

## Additional context
Checkov rates code signing as a high finding: https://docs.prismacloud.io/en/enterprise-edition/policy-reference/aws-policies/aws-general-policies/bc-aws-272

## Breaking Changes
No

## How Has This Been Tested?
- [ X ] I have updated at least one of the `examples/code-signing/main.tf` to demonstrate and validate my change(s)


examples/code-signing/main.tf
```diff
@@ -14,21 +17,23 @@ provider "aws" {
...
 
-  function_name           = random_pet.this.id
-  handler                 = "index.lambda_handler"
-  runtime                 = "python3.12"
-  code_signing_config_arn = aws_lambda_code_signing_config.this.arn
-  create_package          = false
+  function_name                    = random_pet.this.id
+  handler                          = "index.lambda_handler"
+  runtime                          = "python3.12"
+  create_package                   = false
```
antonbabenko, line 23: `create_package = true` should also be supported.

We also need to ensure that the module creates all the AWS Signer resources required for this (conditionally), as this is key to implementing this feature within the module.

variables.tf
```diff
@@ -847,3 +847,37 @@ variable "recursive_loop" {
   type        = string
   default     = null
 }
+
+###############
+# Code Signing
+###############
+
+variable "enable_code_signing" {
+  description = "Must be used with a lambda storing code on s3. Set this to true for triggering a signing job creating a signed copy of the lambda zip. https://docs.aws.amazon.com/lambda/latest/dg/configuration-codesigning.html"
```
antonbabenko, line 856: Please remove links in the descriptions.

antonbabenko (changes requested):
It is a start, but it needs a bit more work to be merged.

examples/code-signing/main.tf
Fran-Rg: Good point. Updated with a packaging happening within.

The following was really to add the signing job as an extra if the conditions are mets (`var.enable_code_signing` + `var.lambda_code_signing_profile_name`). 

Then all arguments on the new added resource are covered: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/signer_signing_job.html

Fran-Rg:
Not stale. @antonbabenko free to review and examples updated

PR #748, opened 2026-05-01
fix: Attempt docker pull before docker build when image not found locally

### LINKED ISSUE #746: docker pull not attempted when docker_image is set but not found locally — falls back to docker build causing failure
## Description

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

When `build_in_docker = true` and `docker_image` is set to a pre-built public registry image (e.g. `public.ecr.aws/sam/build-python3.12:latest`), the module checks whether the image exists locally via `docker images`. If the image is **not present locally** (fresh machine, CI runner, or after `docker system prune`), instead of pulling it from the registry, `package.py` falls back to `docker build` using `docker_build_root` as the build context.

Because `docker_build_root` is typically the Lambda source directory (which contains no `Dockerfile`), the build fails immediately:

```
ERROR: failed to build: failed to solve: failed to read dockerfile: open Dockerfile: no such file or directory
```

The workaround is to manually run `docker pull <image>` before `terraform apply`, but this should not be required.

---

## ⚠️ Note

- Local `.terraform` directory was removed and `terraform init` was re-run before filing this issue.
- Issue persists consistently on a clean Docker image cache.

---

## Versions

- **Module version:** 8.8.0
- **Terraform version:** v1.14.8 (darwin_arm64)
- **Provider versions:**
  - `hashicorp/aws` v6.43.0
  - `hashicorp/null` v3.2.4
  - `hashicorp/local` v2.8.0
  - `hashicorp/external` v2.3.5

---

## Reproduction Code [Required]

```hcl
module "my_lambda" {
  source  = "terraform-aws-modules/lambda/aws"
  version = "~> 8.0"

  function_name = "my-lambda"
  handler       = "main.lambda_handler"
  runtime       = "python3.12"

  create_package = true
  source_path = [{
    path             = "./my_lambda_source"
    pip_requirements = true
  }]

  build_in_docker   = true
  docker_image      = "public.ecr.aws/sam/build-python3.12:latest"
  docker_build_root = "./my_lambda_source"  # contains main.py + requirements.txt, no Dockerfile

  architectures = ["arm64"]
}
```

Where `./my_lambda_source` contains only `main.py` and `requirements.txt` — no `Dockerfile`.

Steps to reproduce:
1. Ensure the image `public.ecr.aws/sam/build-python3.12:latest` is **not** present locally (`docker rmi public.ecr.aws/sam/build-python3.12:latest` or use a fresh environment).
2. Run `terraform init && terraform apply`.
3. Observe failure during the `null_resource.archive` local-exec provisioner.

---

## Expected behavior

When `docker_image` is set and the image is not found locally, `package.py` should run `docker pull <image>` to fetch it from the registry before proceeding with the build.

If the pull of the public image fails for some reason, there is an opportunity to provide a clear error message to the user, instead of an opague no-such-file message.
---

## Actual behavior

`package.py` runs `docker build --tag <image> <docker_build_root>` as the fallback. Since the Lambda source directory has no `Dockerfile`, this immediately fails:

```
> docker images '--format={{.ID}}' public.ecr.aws/sam/build-python3.12:latest
> docker build --tag public.ecr.aws/sam/build-python3.12:latest ./my_lambda_source
ERROR: failed to build: failed to solve: failed to read dockerfile: open Dockerfile: no such file or directory
zip: Error during zip archive creation
subprocess.CalledProcessError: Command '["docker", "build", "--tag", "public.ecr.aws/sam/build-python3.12:latest", "./my_lambda_source"]' returned non-zero exit status 1.
```

---

## Root cause

In `package.py`, function `install_pip_requirements` (~line 1173):

```python
ok = False
while True:
    output = check_output(docker_image_id_command(docker_image))  # docker images --format=...
    if output:
        docker_image_tag_id = output.decode().strip()
        ok = True
    if ok:
        break
    docker_cmd = docker_build_command(   # <-- should be docker pull, not docker build
        build_root=docker_build_root,
        docker_file=docker_file,
        tag=docker_image,
    )
    check_call(docker_cmd)
    ok = True
```

When the image is not cached locally, the fallback unconditionally calls `docker build`. For a custom image with a user-supplied `Dockerfile` this might be intentional, but when no `docker_file` is provided (the standard `docker_image` use case), the correct fallback is `docker pull`.

---

## Suggested fix

When `docker_file` is `None`, fall back to `docker pull` rather than `docker build`:

```python
if docker_file:
    docker_cmd = docker_build_command(
        build_root=docker_build_root,
        docker_file=docker_file,
        tag=docker_image,
    )
else:
    docker_cmd = ["docker", "pull", docker_image]
check_call(docker_cmd)
```

---

## Additional context

- Running on macOS (darwin/arm64) with Colima as the Docker runtime.
- The issue is silent on machines where the image is already cached, which is why it may not have been caught earlier — it only manifests on a clean environment.

raman1236:
## Description

Fixes #746

When `build_in_docker = true` and `docker_image` is set to a pre-built public registry image (e.g. `public.ecr.aws/sam/build-python3.12:latest`), if the image is not present locally, `package.py` falls back directly to `docker build` using `docker_build_root` as the build context. Since Lambda source directories typically don't contain a Dockerfile, this fails with:

```
ERROR: failed to build: failed to solve: failed to read dockerfile: open Dockerfile: no such file or directory
```

## Changes

- Added `docker_pull_command()` helper function alongside existing `docker_image_id_command()` and `docker_build_command()`
- Modified all 4 docker image resolution paths (pip, poetry, npm/yarn, uv) to attempt `docker pull` before falling back to `docker build`
- If `docker pull` fails (e.g. for custom/local images), it falls back to the original `docker build` behavior

## Impact

- Pre-built registry images: `docker pull` succeeds, image is available locally
- Custom images with Dockerfile: `docker pull` fails, falls back to `docker build` (same as before)
- Already-local images: no change (image found in `docker images` check)

### DIFF
--- a/package.py
+++ b/package.py
@@ -1182,12 +1182,15 @@ def install_pip_requirements(query, requirements_file, tmp_dir):
                     ok = True
                 if ok:
                     break
-                docker_cmd = docker_build_command(
-                    build_root=docker_build_root,
-                    docker_file=docker_file,
-                    tag=docker_image,
-                )
-                check_call(docker_cmd)
+                try:
+                    check_call(docker_pull_command(docker_image))
+                except CalledProcessError:
+                    docker_cmd = docker_build_command(
+                        build_root=docker_build_root,
+                        docker_file=docker_file,
+                        tag=docker_image,
+                    )
+                    check_call(docker_cmd)
                 ok = True
         elif docker_file or docker_build_root:
             raise ValueError(
@@ -1325,12 +1328,15 @@ def install_poetry_dependencies(query, path, poetry_export_extra_args, tmp_dir):
                     ok = True
                 if ok:
                     break
-                docker_cmd = docker_build_command(
-                    build_root=docker_build_root,
-                    docker_file=docker_file,
-                    tag=docker_image,
-                )
-                check_call(docker_cmd)
+                try:
+                    check_call(docker_pull_command(docker_image))
+                except CalledProcessError:
+                    docker_cmd = docker_build_command(
+                        build_root=docker_build_root,
+                        docker_file=docker_file,
+                        tag=docker_image,
+                    )
+                    check_call(docker_cmd)
                 ok = True
         elif docker_file or docker_build_root:
             raise ValueError(
@@ -1522,12 +1528,15 @@ def install_uv_dependencies(query, path, uv_export_extra_args, tmp_dir):
                 check_output(docker_image_id_command(docker_image)).decode().strip()
             )
             if not output:
-                docker_cmd = docker_build_command(
-                    build_root=docker_build_root,
-                    docker_file=docker_file,
-                    tag=docker_image,
-                )
-                check_call(docker_cmd)
+                try:
+                    check_call(docker_pull_command(docker_image))
+                except CalledProcessError:
+                    docker_cmd = docker_build_command(
+                        build_root=docker_build_root,
+                        docker_file=docker_file,
+                        tag=docker_image,
+                    )
+                    check_call(docker_cmd)
                 output = (
                     check_output(docker_image_id_command(docker_image)).decode().strip()
                 )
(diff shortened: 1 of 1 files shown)


baolsen:
Not stale, still an issue.

PR #755, opened 2026-05-20
introduced image uri lifecycle

rahul1-hoag:


antonbabenko:
Good in theory but not in reality

PR #761, opened 2026-07-01
feat: Add region variable to alias submodule for provider v6 resource-level region

### LINKED ISSUE #760: modules/alias: missing region attribute for AWS provider v6 resource-level region support
## Summary

AWS provider v6 introduced a resource-level `region` attribute on `aws_lambda_alias`, `aws_lambda_permission`, `data.aws_lambda_alias`, and `aws_lambda_event_source_mapping`, allowing resources to target a specific region without requiring provider aliases.

The `modules/alias` submodule currently has no `region` variable and does not pass `region` to any of its resources. This means callers who set `region` on the root module (via its new `region` variable) cannot guarantee the alias and its associated resources land in the same region — they silently fall back to the provider's configured region.

## Steps to reproduce

1. Configure an AWS provider for `us-west-2`
2. Call the root `terraform-aws-lambda` module with `region = "us-east-2"` — the Lambda function is created in `us-east-2` ✅
3. Call `modules/alias` pointing to that function's ARN — the alias is created in `us-west-2` (provider default) ❌, causing a `ResourceNotFoundException` at apply time because the function doesn't exist in `us-west-2`

## Expected behaviour

`modules/alias` should expose an optional `region` variable (default `null`) and pass it to all regional resources:

```hcl
variable "region" {
  description = "AWS region. Null uses the provider's configured region."
  type        = string
  default     = null
}
```

Applied to:
- `aws_lambda_alias` (both `no_refresh` and `with_refresh`)
- `data "aws_lambda_alias"`
- `aws_lambda_permission` (`version_triggers` and `qualified_alias_triggers`)
- `aws_lambda_function_event_invoke_config`
- `aws_lambda_event_source_mapping`

## Provider version

AWS provider `>= 6.0.0`

## References

- [aws_lambda_alias region argument](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_alias) — confirmed `region` is supported in provider v6
- Root module `region` variable was added to support provider v6 resource-level regions

somaz94:
The `modules/alias` submodule had no `region` variable, so callers relying on the root module's provider-v6 resource-level `region` support could not guarantee the alias and its associated resources land in the same region as the function. They silently fell back to the provider's configured region, causing a `ResourceNotFoundException` at apply time when the function lived in a different region.

This adds an optional `region` variable (default `null`) to `modules/alias` and passes it to all seven regional resources, mirroring the root module's existing pattern:

- `data.aws_lambda_alias.existing`
- `aws_lambda_alias.no_refresh` / `aws_lambda_alias.with_refresh`
- `aws_lambda_function_event_invoke_config.this`
- `aws_lambda_permission.version_triggers` / `aws_lambda_permission.qualified_alias_triggers`
- `aws_lambda_event_source_mapping.this`

The `wrappers/alias` passthrough and the terraform-docs README row are included. Backward compatible: `default = null` is a no-op for existing users, and the provider requirement is already `>= 6.28` so no version bump is needed.

Validation (local, all green):
- `terraform fmt -check -recursive`: clean
- `terraform validate` (init -backend=false): Success
- `tflint`: exit 0
- `terraform-docs` (v0.20.0): README Inputs table in sync, region row added

closes #760

### DIFF
--- a/modules/alias/README.md
+++ b/modules/alias/README.md
@@ -161,6 +161,7 @@ No modules.
 | <a name="input_maximum_retry_attempts"></a> [maximum\_retry\_attempts](#input\_maximum\_retry\_attempts) | Maximum number of times to retry when the function returns an error. Valid values between 0 and 2. Defaults to 2. | `number` | `null` | no |
 | <a name="input_name"></a> [name](#input\_name) | Name for the alias you are creating. | `string` | `""` | no |
 | <a name="input_refresh_alias"></a> [refresh\_alias](#input\_refresh\_alias) | Whether to refresh function version used in the alias. Useful when using this module together with external tool do deployments (eg, AWS CodeDeploy). | `bool` | `true` | no |
+| <a name="input_region"></a> [region](#input\_region) | Region where the resource(s) will be managed. Defaults to the region set in the provider configuration | `string` | `null` | no |
 | <a name="input_routing_additional_version_weights"></a> [routing\_additional\_version\_weights](#input\_routing\_additional\_version\_weights) | A map that defines the proportion of events that should be sent to different versions of a lambda function. | `map(number)` | `{}` | no |
 | <a name="input_use_existing_alias"></a> [use\_existing\_alias](#input\_use\_existing\_alias) | Whether to manage existing alias instead of creating a new one. Useful when using this module together with external tool do deployments (eg, AWS CodeDeploy). | `bool` | `false` | no |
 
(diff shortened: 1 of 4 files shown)


somaz94:
@antonbabenko Still relevant — this adds the `region` variable to the alias submodule so it works with provider v6 resource-level region. CI is green. Could you take a look when you get a chance?

somaz94:
Open since June with no review, and the same goes for #762 alongside it. This one adds a `region` variable to the alias submodule for provider v6 resource-level regions; #762 is a docs clarification that `lambda_role` is ignored when `create_role` is true. Both green. Happy to close either if they are not wanted.

Issue #744, opened 2026-04-22
lambda_role ignored by default


rgarrigue:
## Description

I broke my production bumping from v7 to v8 as I didn't saw that lambda_role is ignored now, due to

>   role                               = var.create_role ? aws_iam_role.lambda[0].arn : var.lambda_role

That's my bad for not paying attention. And my bad for not strong-arming my dev to have a dev env for this event if they don't want to. 

But maybe could be documented a bit more, like in the lambda_role description, or if possible an assert to block the combination of create_role = true && lambda_role != ""

## Versions

- Module version [Required]: 8.7.0

rgarrigue:
Not stale

PR #762, opened 2026-07-03
docs: Clarify lambda_role is ignored when create_role is true

somaz94:
The lambda_role variable is silently ignored when create_role is left at its
default (true), because role is wired as
`var.create_role ? aws_iam_role.lambda[0].arn : var.lambda_role`. A user hit
this on a v7 -> v8 bump and broke production (#744).

This clarifies the lambda_role description so its interaction with create_role
is explicit. It also drops a stray leading space and fixes an "our" -> "your"
typo in the same sentence.

Validation:
- terraform fmt -check: clean
- README inputs table row regenerated to match (terraform-docs inject; escaped
  underscores and the existing table-separator style preserved to avoid v0.24
  formatting drift)

related: #744


somaz94:
@antonbabenko Still relevant — small docs clarification, no code change. CI is green. Would appreciate a look when you have time.

somaz94:
Still relevant. The wiring this documents is unchanged on master: main.tf line 48 is still `role = var.create_role ? aws_iam_role.lambda[0].arn : var.lambda_role`, so lambda_role is still silently ignored at the default create_role = true.

Worth noting that #744, the issue that prompted this, was closed on 2026-07-04 without anything being added to the docs, so the next person hits the same thing. That is all this change is: 2 lines across variables.tf and README.md, no code change, mergeable, CI green.

Happy to close it if you would rather not carry the wording.

PR #770, opened 2026-08-31
feat: Enabling poller_group_name atribut for reducing Event Poller Unit lambda costs

sramekmichal:
## Description
Add poller_group_name pass-through to the provisioned_poller_config dynamic block in aws_lambda_event_source_mapping resource.
  
Currently the module only passes minimum_pollers and maximum_pollers but omits poller_group_name, which is supported by the AWS provider since v5.62.0.

## Motivation and Context
AWS Lambda supports ESM grouping via poller_group_name in provisioned_poller_config. This allows multiple event source mappings within the same VPC to share Event Poller Unit (EPU) capacity, reducing Provisioned mode costs by up to 90% for low-throughput workloads.
  
Without this parameter, users who need poller_group_name must either fork the module or manage event source mappings outside of it.
  
Reference: [Link](https://docs.aws.amazon.com/lambda/latest/dg/kafka-scaling-modes.html)

## Breaking Changes
None. The parameter defaults to null via try(), so existing configurations are unaffected.

## How Has This Been Tested?
  - [x] I have updated at least one of the examples/* to demonstrate and validate my change(s)
  - [x] I have tested and validated these changes using one or more of the provided examples/* projects
  - [x] I have executed pre-commit run -a on my pull request


sramekmichal:
@antonbabenko Very relevant and helpful, check it out, how I saved daily costs for the EPU capacity after deploying provided changes:
image

sramekmichal:
@bryantbiggs

sramekmichal:
@antonbabenko ?

antonbabenko (commented):
**Verdict: changes suggested.**

One MEDIUM finding is open (ladder rule 6).

### Findings

**MEDIUM**

- `examples.feature-undemonstrated` - `examples/event-source-mapping` - No example sets `provisioned_poller_config`, so nothing in `examples/` exercises the new `poller_group_name` argument or the block it belongs to. The change touches only `main.tf`, although the description says an example was updated.

---

Reviewed against the terraform-aws-modules profile at `ea68395`.

sramekmichal:
Thanks! True, updated example.

antonbabenko (changes requested):
The module change is right: `poller_group_name` is passed through the existing `provisioned_poller_config` block, and provider 6.28.0 (the module's current floor) already supports it, so no version bump is needed. The updated example is the problem: the new `apache_kafka` mapping has no event source at all.

**Changes requested before merge.**

A HIGH finding is open (ladder rule 2).

### Findings

**HIGH**

- `examples.example-broken` - `examples/event-source-mapping/main.tf:92` - The `apache_kafka` entry sets neither `event_source_arn` nor `self_managed_event_source` (the latter is commented out), so the mapping has no source and `CreateEventSourceMapping` rejects it at apply; `provisioned_poller_config` is only valid for MSK or self-managed Kafka sources. `terraform validate` passes, so this was found by reading the provider schema, not by a plan or apply. Fix by giving the entry a real Kafka source (an MSK cluster in the example's existing VPC, or a live `self_managed_event_source` with its `source_access_configuration`). Commenting the entry back out is not a fix, since the new argument would again be demonstrated nowhere.

antonbabenko (changes requested):
The `poller_group_name` pass-through is right, and the updated example now gives the Kafka mapping a real self-managed source, which fixes the earlier gap. One value in that example will still be rejected at apply. `terraform validate` and `terraform plan` both pass on the example.

**Changes requested before merge.**

A HIGH finding is open (ladder rule 2).

### Findings

**HIGH**

- `examples.example-broken` - `examples/event-source-mapping/main.tf:95` - The `apache_kafka` mapping sets `topics = ["topic1", "topic2"]`, but the provider schema at the module's floor (6.28.0) says a single topic name must be specified, so `CreateEventSourceMapping` rejects the list at apply. `terraform plan` passes, because the provider does not check the list length at plan time, so this comes from the schema. Fix by setting one topic, for example `topics = ["topic1"]`.

### Pull request state

Checks pass. Mergeable. Review decision: changes requested (earlier review). No open threads.

antonbabenko (approved):
The change passes `poller_group_name` through the existing `provisioned_poller_config` block, and the module's provider floor (6.28) already has the argument. The example now sets a single topic, which fixes the finding from the last review. `terraform validate` passes on `examples/event-source-mapping`; `terraform plan` was not run this time.

**Good to merge.**

No findings, and every pull request signal passes (ladder rule 8).

### Pull request state

Earlier review from this account: changes requested on `da93dc8`; this review replaces it.

antonbabenko:
This PR is included in version 8.9.0 :tada:

